PR #392, opened 2014-12-08
Add Cond example

zryty:


krakjoe:
This code isn't really of a good enough standard.

Using globals is bad, refactor to inject those deps.
Using mutex and cond directly is difficult, which is why normal users shouldn't do it. This code is vulnerable to spurious wakeups and includes advice which is bad (omitting lock/unlock when using timeout).
It's preferred that the programmer use the higher level ::synchronized, ::wait and ::notify to achieve the same thing.

I don't object to having a good example for Condition and Mutex, but this isn't it, feel free to have another go and thanks for taking the time ;)

zryty:
locking/unlocking mutexes not work so well in php, many times script hangs on it and as of "php pthreads carrying about protect variables" we don't need them anymore.

spurious wakeups - add chcecking if it is a new number. Even examples on C++ have something like:
```
while(data_count > 0)
     cond_wait(...); // something like that
```

globals - this isn't point of example. "First done thing, then optimize it" ;)

Better example I won't write, because I'm done script with this extension and will take some time before I will write another script with it. Anyway If someone want to find something about "conditional variables" maybe find this request.

Thanks for your time ;)

PR #407, opened 2015-01-02
PTHREADS_ZG(resources) destroyed too early

2072:
Finally fixed the issue uncovered by working on #404, the fix was actually simple:

PTHREADS_ZG(resources) was destroyed before destroying EG(symbol_table): resources marked for keeping would be destroyed if stored in local variables inside ::run() and not destroyed in user-land before the end of the process.

This is fixed by declaring the PTHREADS_ZG(resources) hash-table as persistent and destroying it in ~~PHP_MSHUTDOWN_FUNCTION~~ ZEND_MODULE_POST_ZEND_DEACTIVATE_D  instead of in PHP_RSHUTDOWN_FUNCTION. Since PTHREADS_ZG(resources) is necessarily left empty at the end of a request, there is no memory waste.

I also added nice tests for this issue (3704fa9) and the previous one (2468f16).

Another thing I like to do is make Pthreads throw an exception if the user tries to explicitly close a foreign resource because doing that now will cause a segmentation fault when the resource is re-accessed in its owner...

### DIFF
--- a/php_pthreads.c
+++ b/php_pthreads.c
@@ -306,7 +306,13 @@ PHP_MSHUTDOWN_FUNCTION(pthreads)
 	if (pthreads_instance == TSRMLS_C) {
 		pthreads_globals_shutdown(TSRMLS_C);
 	}
-	
+
+	if (PTHREADS_ZG(resources)) {
+		zend_hash_destroy(PTHREADS_ZG(resources));
+		FREE_HASHTABLE(PTHREADS_ZG(resources));
+		PTHREADS_ZG(resources) = NULL;
+	}
+
 	return SUCCESS;
 }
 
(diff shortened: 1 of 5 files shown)


2072:
needs more testing just got a segfault in my application when a new worker was started...
on `if (PTHREADS_ZG(resources) && zend_hash_find(PTHREADS_ZG(resources)` right at the beginning of a `worker::run()` where no foreign resource could have been added to `PTHREADS_ZG(resources)`... so PTHREADS_ZG(resources) should have been null but it was not which makes no sense to me... or maybe a compiler optimization got in the way?

I'm letting a few instances of my app run for a day and I'll reopen this when I'm sure everything is fine. 
I'm still discovering PHP extension coding so there might be an obvious mistake in what I did...

2072:
ok well the problem was that PHP_MSHUTDOWN_FUNCTION was only called once when the main thread exits while PHP_MINIT_FUNCTION does get called every time a thread starts... weird.

so I moved PTHREADS_ZG(resources) into a new callback ZEND_MODULE_POST_ZEND_DEACTIVATE_D which does get called appropriately and seems the most apporpriate place to do that according to http://php.net/manual/fr/internals2.structure.modstruct.php

It should be ok now.

krakjoe:
Very nice, thanks :) Moar internet points for you :100:

PR #419, opened 2015-03-09
fixes win-32 seqfault when system resources limit reached

zelgerj:


krakjoe:
Presumably, you have debugged the problem and found that the crash occurred in ```pthread_cond_destroy```, and deduced that sync must have been ```NULL``` when ```pthread_cond_destroy``` was called, so using a predicate on ```sync``` should work.

If sync really was ```NULL```, the call to lock free and free underneath the call to ```pthread_cond_destroy``` would also crash.

I think you just moved the crash, but it will still happen, maybe a line further down or on another line entirely.

The resource limit being reached will be detectable at certain pthread_ calls. It's probably the case that one of the times I am using a pthread_ function, I am not checking the return value, the proper thing to do is manage the problem at the point of origin, not work around the side effects of the problem (such as random pointers appearing null).

zelgerj:
Hi Joe,

thank you for that fast response. You're absolutely right, but the code a line further down will not crash at all... I'am also interested in fixing the problem at the point of origin but it was not able for me to find it as it seems to be a raise condition, because there is still the same problem in `pthreads_synchro_alloc` which calls `pthread_cond_init($sync->notify, NULL)` that also crashes from time to time while having a dead pointer... :(

The message *"pthreads has detected that the %s could not be started, the system lacks the necessary resources or the system-imposed limit would be exceeded"* that will normally be thrown in case of an EAGAIN only comes out if i supply that simple fix given in this pull-request...

I tried to create a simple test script for that problem but yet without success...
Maybe you have any clue?

Cheers!
Johann

zelgerj:
I just reseted my master so the pull-request has beed closed, lets discuss that using an issue.

PR #422, opened 2015-03-09
fixes #362, massive performance loss

### LINKED ISSUE #362: Massive performance loss
Hi Joe,

as in Issue #361 described, we're actually try to update to latest pthreads 2.0.10. While we still have some problems with pthreads > 2.0.8 version 2.0.7 is working now :)

After some performance benchmarks we see a massive performance loss with 2.0.7 documented by the following screenshots:

pthreads-2 0 7
pthreads-2 0 10

I think that a performance impact around 5 % because of improved stability, security and memory handling will be quite o. k., but a slow down from 234 req/s to  2,33 req/s when calling an example servlet is quite too massive. and makes the appserver absolutely obsolete.

As i'm not sure, if the problem is based in our code, it would be helpful, if you can specify a list of things we should ignore to optimise performance. On the other hand, i think it can't be in our code, because 1.0.1 works almost perfect.

If you want to test it by yourself, you can simple download the latest appserver.io version from builds.appserver.io and install pthreads 2.0.7.

Thanks
Tim

zelgerj:


krakjoe:
This causes tests to fail on travis.

PR #425, opened 2015-03-12
Fix possible race condition in tests/binding.phpt as seen in #424

2072:
`::notify()` calls were not synchronized and so in the following code:

```php
foreach($threads as $thread) {
	$thread->synchronized(function() use($thread){
		if (!$thread->done)
			var_dump($thread->wait());
		else var_dump($thread->done);
	});
}
```

`$thread->done()` could be set to true just between the `if` and the `::wait()` if the treads are started faster than usual as is expected with #424.

### DIFF
--- a/tests/binding.phpt
+++ b/tests/binding.phpt
@@ -12,8 +12,10 @@ class ThreadTesting extends Thread {
 		$this->other = $other;
 	}
 	public function run(){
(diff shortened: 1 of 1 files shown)


zelgerj:
You're damn right! :) Good one!

PR #446, opened 2015-06-23
fixed inherit globals

zelgerj:
* fixed inherit globals to not inherit creator ls in favour of global ls


zelgerj:
problem is not fixed... il'll post an issue for that with an code example...

PR #448, opened 2015-06-29
added test for globals-require bug

zelgerj:


zelgerj:
Hi Joe, would be cool if you might have a look at it... Cheers!

2072:
Don't forget to include your test in package.xml and maybe you could use --XFAIL-- to make Travis happy?

krakjoe:
Not applicable to seven, so closing.

Issue #441, opened 2015-04-26
Pool does not free work on collection


TheUbuntuGuy:
I have submitted this before and it was dismissed as not a bug. I apologize for bringing this up again, but I don't think I got across what was actually happening very well, and I cannot see how the behaviour I am seeing is intended. I have tried to patch this myself, however not being a PHP internals developer, I did not get anywhere productive. All the tests I did were using PHP 5.6.8 and the latest source of pthreads.

From what I understand, once pool->collect() is called and the function returns true, then the associated work object should be subsequently destroyed and freed. However this is not what I am seeing. I modified the pooling.php example to better illustrate this.

```
<?php
class WebWorker extends Worker {

	public function __construct() {
		echo "constructing worker\n";
	}
	
	public function __destruct() {
	        echo "destructing worker\n";
	}
}

class WebWork extends Collectable {
	public function run() {
		$this->setGarbage();
	}
	
	public function __destruct() {
	        echo "destructing\n";
	}
}

echo "Creating pool...\n";
$pool = new Pool(4, 'WebWorker');

echo "Submitting Work...\n";
for ($i=0;$i<10;$i++) {
        $pool->submit(new WebWork());
}

echo "Collecting from Pool...\n";
$pool->collect(function($work){
        echo "collecting\n";
	return $work->isGarbage();
});

echo "Shutting Down Pool...\n";
$pool->shutdown();

var_dump($pool);

unset($pool);

echo "End of Program...\n";
?>
```

This is pretty straightforward: create a pool, submit some work, and collect the results. However when running this you will find that the WebWork objects are not being destroyed until the end of the program, instead of after they are collected.

```
Creating pool...
Submitting Work...
constructing worker
constructing worker
constructing worker
constructing worker
Collecting from Pool...
collecting
collecting
collecting
collecting
collecting
collecting
collecting
collecting
collecting
collecting
Shutting Down Pool...
class Pool#1 (6) {
  protected $size =>
  int(4)
  protected $class =>
  string(9) "WebWorker"
  protected $workers =>
  array(0) {
  }
  protected $work =>
  array(0) {
  }
  protected $ctor =>
  NULL
  protected $last =>
  int(2)
}
End of Program...
destructing
destructing worker
destructing
destructing worker
destructing
destructing worker
destructing
destructing worker
destructing
destructing
destructing
destructing
destructing
destructing
```

[...]

PR #454, opened 2015-07-09
Fix #452: Call user-land ::__destruct() before freeing a threaded innards

### LINKED ISSUE #452: __pthread_mutex_lock "assertion failed" when called from __destruct()
I switched to the current pthreads-master due to a traits problem I was experiencing in 2.0.10.

Now I ran into another problem:
```php
class Work extends Threaded
{
    public function __destruct()
    {
        var_dump(self::isRunning());
    }
}
$work = new Work;
```
Result:
> php: ../nptl/pthread_mutex_lock.c:80: __pthread_mutex_lock: Assertion `mutex->__data.__owner == 0' failed.
> Aborted

When the object is freed with `unset($work)` before script shutdown, it works.

Bug? No Bug? Some configure flag that needs to be set?

*Using PHP-CLI 5.6.10 under Linux*

2072:
So that threaded members can be accessed as usual inside ::__destruct()

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -89,6 +89,7 @@
    <file name="tests/null-member-crash.phpt" role="test" />
    <file name="tests/oomethods.phpt" role="test" />
    <file name="tests/objects-from-closures.phpt" role="test" />
+   <file name="tests/orderly-destruction.phpt" role="test" />
    <file name="tests/pools-default.phpt" role="test" />
    <file name="tests/pools.phpt" role="test" />
    <file name="tests/resource-keeping.phpt" role="test" />
--- a/src/object.c
+++ b/src/object.c
@@ -580,47 +580,71 @@ static void pthreads_base_ctor(PTHREAD base, zend_class_entry *entry TSRMLS_DC) 
 static void pthreads_base_dtor(void *arg, zend_object_handle handle TSRMLS_DC) {
     PTHREAD base = (PTHREAD) arg;
 
-	if (PTHREADS_IS_NOT_CONNECTION(base) && PTHREADS_IS_NOT_DETACHED(base)) {
+	if (PTHREADS_IS_NOT_DETACHED(base)) {
 	    assert(base->cls == TSRMLS_C);
 
-	    if (PTHREADS_IS_THREAD(base)||PTHREADS_IS_WORKER(base)) {
-	        pthread_t *pthread = &base->thread;
-	        if (pthread) {
-		        pthreads_join(base TSRMLS_CC);
-	        }
-        }
-	    // call userland destructor if any
-	    zend_objects_destroy_object(arg, handle TSRMLS_CC);
+            if (PTHREADS_IS_NOT_CONNECTION(base)) {
+                if (PTHREADS_IS_THREAD(base)||PTHREADS_IS_WORKER(base)) {
+                    pthread_t *pthread = &base->thread;
+                    if (pthread) {
+                        pthreads_join(base TSRMLS_CC);
+                    }
+                }
+                // call userland destructor if any
+                zend_objects_destroy_object(arg, handle TSRMLS_CC);
 
-	    pthreads_lock_free(base->lock TSRMLS_CC);
-	    pthreads_state_free(base->state  TSRMLS_CC);
-	    pthreads_modifiers_free(base->modifiers TSRMLS_CC);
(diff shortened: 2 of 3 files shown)


shoghicp:
Will this also free the Threaded object itself? There were memory leaks with them

2072:
No these are two different issues, I've just tested #441 to make sure and it's still happening, but this leak issue only happens if one uses the provided pool implementation which add references to submitted task but never removes them so that they are only removed when the thread containing the pool terminates.

shoghicp:
@2072 I implemented a simple version of pool, and while it's way less leaks than before, something still leaks there :(

Thanks for your fix, this also fixes some issues we were having when stopping the server!

2072:
@shoghicp: Then this is yet another issue, you should open a new one with code to reproduce it along with your pool implementation.

As for #441, I checked `pool.h` code and it's clear that references to submitted tasks are not removed upon collection. It's probably a one line fix but my understanding of PHP's internal reference management is still scarce so I prefer to let @krakjoe fix that one...

krakjoe:
@2072 I see you're very active, appreciate that.

I am extremely busy with PHP7, I don't want to change my focus right now.

So, this is definitely a valid fix, however, there is another mistake (my mistake). The dtor does need to be called for connection objects too, it can't be called for a detached object though (::detach is going to be removed for 7, it's too dodgy).

This does account for a rise in memory usage over time, leaks, but of the kind that may not be reported by zend because the gc probably clears them up for you on process exit, so no need to report them.

This is fixed in 7, if you want to change and test your patch I'll happily merge it into the 5 branch.

I've found a few such mistakes and fixed them in the seven branch, and will backport changes, where possible, when I'm done working on the new branch.

There will always be a branch for 5 and 7, the code is so different that there has to be. Because porting to 7 represents a huge amount of work going forward, when the stable releases of PHP7 come out, I will be dropping active support for the 5 branch (once the aforementioned fixes are merged and a final release is tagged).

I realise there are still people using 5 and will be at that time, but other than the help I get from PR's, I'm all on my own still, years later. So, this is a necessary move.

2072:
@shoghicp could you test if that fixes your leak? (still by using your own pool implementation)

@krakjoe as I said in my last commit I partly reverted f4ef9c8 the main difference is the alternate version of  zend_object_std_dtor used when `+#if PHP_VERSION_ID > 50399` which seems to  destroy more thoroughly an object's properties table. Apart from that, connection objects did fall through the same destruction process as regular pthreads objects (as far as I could see).

While trying to understand how these connection objects were used I discovered another problem with `PTHREADS_ZG(cache)` hashtable which is used to store addresses of 'unserialized' PTHREAD object: The addresses added there are never removed which can cause addresses collisions (as well as a memory leak) if a shared PTHREAD is deleted in its original context.

I've actually experienced this problem in my own software some time ago while trying to share objects between threads - because of this problem shared PTHREAD objects cannot be deleted but all is well if you 'recycle' them by emptying them and reconstructing them when you need a new one...

smolinari:
@krakjoe Hey Joe. Just to be certain i am understanding your post above correctly. You are working on a branch of pthreads, which will be PHP7 ready? :smiley: 

Scott

2072:
@smolinari check https://twitter.com/krakjoe

cottton:
what i am supposed to do to get it fixed?
I tried a lot and read a lot but i dont get it -- is it possible to get this fixed using the latest version?
```php
class Task extends Collectable
{
    public function __destruct()
    {
// ?
    }
}
class MyWorker extends Worker
{
    public function __destruct()
    {
// ?
    }
}
$Pool = new Pool(10, 'MyWorker');
while(1){
    $Task = new Task();
    $Pool->submit($Task);
    $Pool->collect(
        function($Task){
            $Task->setGarbage();
            return $Task->isGarbage();
        }
    );
    echo round(memory_get_usage(true) / 1048576, 2). ' MB' . PHP_EOL;
    // growing mem until fatal err
}
```

mkmanwich:
@cottton 
This issue doesn't exist on pthreads 2.0.7!!
Try downgrading and running it again, but change the collectable class to threaded (as it doesn't exist in that version).
you will notice 2 things:
1. If you collect, memory doesn't grow.
2. Destructors are never called, not even once, for workers AND for threaded objects.

Something in the new releases messed things up....

krakjoe:
Not applicable to seven, so closing ...

PR #477, opened 2015-08-27
Fix #391 for PHP 5.x

### LINKED ISSUE #391: Saving an object in a static scope variable results into an invalid pointer
to save an object in a static scope results into an invalid pointer.

See following example:
```php
<?php

class AbstractClass extends Stackable  { }

class A extends AbstractClass {
	protected $numberX;
	
	public function setNumberX($x) { $this->numberX = $x; }
	public function getNumberX() { return $this->numberX; }
}

class B extends AbstractClass {
	protected $numberX;
	protected $numberAuditTime;
    protected $objectA;
	
	public function setObjectA($a) { $this->objectA = $a; }
	public function getObjectA()	{		return $this->objectA;	}
	public function setNumberX($x)	{		$this->numberX = $x;	}
	public function getNumberX()	{		return $this->numberX;	}
	public function getNumberAuditTime()	{		return $this->numberAuditTime;	}
}

function dosomething1() // this fails
{
	static $a=null;
	if($a==null) $a = new A();
	$list = array();
	
	for($i=0;$i<10;$i++)
	{
		//$a->setNumberX(rand(0,50)); // modify it
		$list[] = $a;
		$b = new B();
		$b->setNumberX(rand(0,50));
		$b->setObjectA($a);
		$list[] = $b;
	}
	return $list;
}

$a = null;
function dosomething2() // this works
{
	global $a;
	if($a==null) $a = new A();
	$list = array();
	
	for($i=0;$i<10;$i++)
	{
		$a->setNumberX(rand(0,50)); // modify it
		$list[] = $a;
		$b = new B();
		$b->setNumberX(rand(0,50));
		$b->setObjectA($a);
		$list[] = $b;
	}
	return $list;
}

class TestThread extends Collectable
{
	private $l;
	public function __construct($list)
	{
		$this->l = $list;
	}
	
	public function run() 
	{
		//usleep(rand(500,100000));
		$l = $this->l;
		foreach($l as $o)
		{
			if(is_object($o)) echo get_class($o).";";
			else { echo "NO OBJECT?"; var_dump($o); }
		}
		$this->setGarbage();
	}
}

$pool = new Pool(10,'Worker'); 
//$pool = new Pool(4,'Worker'); 
for($i=0;$i<10;$i++)
{
	$all[$i] = dosomething1(); // fails
	//$all[$i] = dosomething2(); // works
	$pool->submit(new TestThread($all[$i]));
	echo "good ".$i."\n";	
}

$pool->shutdown();

die("done\n");
```

with dosomethign1() I see following debug outputs: 
```
/home/adminuser/Downloads/php-5.6.3/Zend/zend_variables.c(188) : Block 0x7ffff7f25308 status:
/home/adminuser/Downloads/php-5.6.3/Zend/zend_execute.h(80) : Actual location (location was relayed)
Invalid pointer: ((thread_id=0xF7FD8600) != (expected=0xEA7FC700))
```

with dosomething2() no invalid pointers occur. The object should still exist though....

regards
Simon

coderstephen:
An attempt at fixing #391 for PHP versions 5.x.

As per the pthreads 2.0.x docs, resources and objects should be nullified inside static variables, so now `pthreads_copy_statics()` skips over those types of values when copying lexical statics by simple type checking. The entry for the variable will instead be empty, which will return a fresh `NULL` after copy to a child thread. Normal zvals are copied and their refcounts incremented.

### DIFF
--- a/src/copy.h
+++ b/src/copy.h
@@ -41,7 +41,7 @@ static HashTable* pthreads_copy_statics(HashTable *old) {
 		) {
 			zend_hash_get_current_key_ex(old, &key_str, &key_str_length, &num_index, 0, &pos);
 
-			if (Z_TYPE_P(*value) != IS_OBJECT && Z_TYPE_P(*value) != IS_RESOURCE) {
+			if (Z_TYPE_PP(value) != IS_ARRAY && Z_TYPE_PP(value) != IS_OBJECT && Z_TYPE_PP(value) != IS_RESOURCE) {
 				zval *copy;
 				ALLOC_ZVAL(copy);
 				MAKE_COPY_ZVAL(value, copy);
--- a/src/copy.h
+++ b/src/copy.h
@@ -21,20 +21,36 @@
 /* {{{ */
 static HashTable* pthreads_copy_statics(HashTable *old) {
 	HashTable *statics = NULL;
-	
+
 	if (old) {
-		zval *tmp;
-	
+		HashPosition pos;
+		zval **value;
+
+		char *key_str;
+		uint key_str_length;
+		ulong num_index;
+
 		ALLOC_HASHTABLE(statics);
 		zend_hash_init(statics,
-			zend_hash_num_elements(old), 
+			zend_hash_num_elements(old),
 			NULL, ZVAL_PTR_DTOR, 0);
-		zend_hash_copy(
-			statics, 
-			old, (copy_ctor_func_t) zval_add_ref, 
-			(void*)&tmp, sizeof(zval*));
+
+		for (zend_hash_internal_pointer_reset_ex(old, &pos);
+			 zend_hash_get_current_data_ex(old, (void **) &value, &pos) == SUCCESS;
+			 zend_hash_move_forward_ex(old, &pos)
+		) {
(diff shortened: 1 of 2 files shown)


krakjoe:
You need to omit copying arrays in the same way for this to work as intended.

Z_TYPE_P(* can be replaced with Z_TYPE_PP in 5

Thanks for taking the time ;)

coderstephen:
OK, will do. Thanks for the tip!

krakjoe:
Excellent work, thanks again @coderstephen 

Sorry I can't find more time to work on the 5 branch, there are a couple of other outstanding issues, when seven is out the door, we'll try and get some of them cleared up for a final release on pecl for the 5 series.

coderstephen:
Glad I can help! Don't feel bad -- it sounds like you've got enough on your plate already.

The only reason I'm so eager for the 5 branch (:smile:) is because I'm simultaneously working on [icicleio/concurrent](https://github.com/icicleio/concurrent) to be 5.x compatible primarily. And I think we're doing some new, unique types of things with pthreads that uncover bugs no one else even noticed. I'm really looking forward to becoming 7 compatible too when pthreads 3 is released.

PR #510, opened 2015-10-30
- some minor fixes

SirSnyder:
- changed pthread.h to php_pthreads.h
- removed check of not-existent sched.h


krakjoe:
That's not right, it needs to check for pthread.h system header.

SirSnyder:
ok, than my include path is wrong

PR #526, opened 2015-11-28
Issue/516

SirSnyder:
Hi @krakjoe, the fix works very well in my tests. Any concerns?

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -243,6 +243,12 @@ int pthreads_connect(pthreads_object_t* source, pthreads_object_t* destination) 
 			return pthreads_connect(source, destination);
 		}
 
+		if(source->originObj) {
+                    destination->originObj = source->originObj;
+                } else {
+                    destination->originObj = (long unsigned int)source;
+                }
+                destination->sourceObj = (long unsigned int)source;
 		destination->thread = source->thread;
 		destination->local.id = source->local.id;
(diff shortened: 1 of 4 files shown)


krakjoe:
Makes every property += 2*sizeof(zend_ulong), which isn't really acceptable ...

Here:

```php
<?php

$objGlobalShare = new Threaded();
$objInstance    = new Threaded();

class TestOne extends Thread
{
    private $objGlobalShare;
    private $objInstance;

    public function __construct(Threaded $globalShare, Threaded $instance)
    {
        $this->objGlobalShare   = $globalShare;
        $this->objInstance      = $instance;
    }

    public function run()
    {
        $this->objGlobalShare['anykey'] = (object) $this->objInstance;
    }
}

$objTestOne = new TestOne($objGlobalShare, $objInstance);
$objTestOne->start();
$objTestOne->join();

$objWaitMutex = new Threaded();
$objWaitMutex->synchronized(function() use ($objWaitMutex, $objGlobalShare, $objInstance)
{
    while(true)
    {
        var_dump($objGlobalShare);

        $objWaitMutex->wait(100000);
    }
});
```

You can use explicit casts to work around the problem and it doesn't cost all your other properties memory.

SirSnyder:
args, should have read object-cast.phpt

PR #609, opened 2016-06-29
Windows Installation Error

stefano-93:
In Windows, the dll pthreadVC2.dll must be put into the Apache\Bin folder in order to work, otherwise the php extension will not be loaded.
Tip found here: http://stackoverflow.com/a/36160612/5260309
Added the step needed to enable the php extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,9 @@ Releases for Windows can be found: http://windows.php.net/downloads/pecl/release
 
 ##### Simple Windows Installation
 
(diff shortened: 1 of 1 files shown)


krakjoe:
Apache has nothing to do with PHP.

stefano-93:
I know, but it works instead doing as you wrote doesn't work. Make a try if you don't trust...
My platform: Windows 8.1 x64, Apache 2.4, PHP 5.6 thread safe

PR #634, opened 2016-11-10
Refactor

tpunt:
Hi Joe!

I've done a little refactoring on parts of pthreads. Nothing seems to be broken by these changes, which focus on removing redundant logic and centralising logic to prevent repetition. All changes are self-contained in individual commits, so if there are any changes you disagree with, then the others can simply be cherry-picked instead.

PR #635, opened 2016-11-11
Fix memory leak

tpunt:
Here's the fix, as mentioned in #634.


krakjoe:
Ah now no-debug faults, I'll have to have a closer look ...

tpunt:
Well that's just irritating ^^. I didn't have much time to look deeper into this today, but I may be able to over the weekend, if you don't find a fix first :).

tpunt:
I haven't managed to make much progress on this, so I'll close this for now.

PR #645, opened 2016-12-12
Synchronized queue functionality

cmyker:
First of all thanks you Joe for making it possible.
I've already developed an experimental application, use it on production, works great! 
But what was quite hard to implement is an ability to simply exchange the data between threads and main, or between different threads.
Mainly it is hard to maintain a refcount to the data you need to pass to the other threads and especially if you have a lot of data to exchange. Good example would be that several threads are asynchronously querying to some web service and main is processing the results (see Queue.php in examples folder).
What I was particularly missing is something similar to python's Queue https://docs.python.org/2/library/queue.html

Keeping in mind that each thread has its own memory manager and the data that thread is passed to other threads or main is allocated dynamically inside that thread, and will be gone after thread is finished or the reference is gone, I came up to the simple solution: 
- store data out of zend memory manager, in the regular heap
- serialize the data that is pushed to the queue, copy that data to the memory allocated in heap, clean up memory that was allocated in zend_mm for serialization
- unserialize data when other thread is poping it out (thus after that it is in the thread's address space), clean up memory allocated in heap
- make it thread safe, thus easy to use in userland, take care of synchronization in C code 
- on object destruction free allocated in heap heap memory if something is left

I was also trying to maintain your code style and stick to the similar design.

### DIFF
--- a/classes/queue.h
+++ b/classes/queue.h
@@ -1,0 +1,101 @@
+/*
+  +----------------------------------------------------------------------+
+  | pthreads                                                             |
+  +----------------------------------------------------------------------+
+  | Copyright (c) Joe Watkins 2012 - 2015                                |
+  +----------------------------------------------------------------------+
+  | This source file is subject to version 3.01 of the PHP license,      |
+  | that is bundled with this package in the file LICENSE, and is        |
+  | available through the world-wide-web at the following url:           |
+  | http://www.php.net/license/3_01.txt                                  |
+  | If you did not receive a copy of the PHP license and are unable to   |
+  | obtain it through the world-wide-web, please send a note to          |
+  | license@php.net so we can mail you a copy immediately.               |
+  +----------------------------------------------------------------------+
+  | Author: Joe Watkins <krakjoe@php.net>                                |
+  +----------------------------------------------------------------------+
+ */
+#ifndef HAVE_PTHREADS_CLASS_QUEUE_H
+#define HAVE_PTHREADS_CLASS_QUEUE_H
+PHP_METHOD(Queue, push);
+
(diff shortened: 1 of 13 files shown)


krakjoe:
This looks quite nice.

I haven't built it yet, will have time in the coming days to review it properly ... I have had a look over it though, follows are some thoughts ...

I'm a bit confused as to how this is different to the queue functionality built into Threaded objects, perhaps you could explain that ?

I can see that it's simpler in some ways, but it still seems to duplicate some functionality, and I'm just not sure why, perhaps I'm missing something obvious ?

pthreads doesn't serialize scalars, it's too wasteful.

There's not enough tests, there needs to be tests that actually use threads too.

cmyker:
> I'm a bit confused as to how this is different to the queue functionality built into Threaded objects, perhaps you could explain that ?

Sorry my bad, now watching closer to the store functionality I see that actually I did is the same (((, except Queue takes care of synchronization
Why I did it in the first place, I was experiencing frequent segfaults, and switching data exchange to the external queue engine (Redis) eliminated those, that is why I thought it would be nice to have such a queue built-in, with no external dependencies...
Does it makes any sense to improve it to the point that you would accept it? Or that is a pure duplicate and don't worth to continue with?

cmyker:
Generally it would be nice to have such a class as python and ruby have... More obvious since this is a very common and also a possibility to provide blocking queue API in the future

krakjoe:
I'm going to say it doesn't really make sense to merge this, sorry about that.

Please do open bug reports that produce the faults you mentioned, and I appreciate you taking the time and effort to try something.

Issue #610, opened 2016-06-29
PHP Crash on Windows 10 when assigning arrays to \Thread


webdes87:
*If you know how to write phpt tests, please open a PR with environment information, and your test, this will cause CI to test your issue before a human is able to look*

Environment
==========

PHP:
D:\dev\xampp\htdocs\musico>php -v
PHP 7.0.5 (cli) (built: Mar 30 2016 10:15:26) ( ZTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies

 OS: WIndows 10 /  x64
Latest pthread release

Summary
======

When using a Thread (class extends \Thread) and assigning an array (propably any complex datatype) to a member variable, php.exe crashes. 
![crash](https://cloud.githubusercontent.com/assets/20090395/16460590/6694846e-3e28-11e6-9acb-108ca5e9bc49.png)

Reproducing Code
==============
https://gist.github.com/webdes87/c90b535f3aa9e72f0d7fa150f104516e

Code will produce error: 

Expected Output
=============
After 1.s Seconds
10 arrays: ["blahkeks", someThreadId]
+ Some progress infos

Actual Output
============
Fatal error: Uncaught RuntimeException: pthreads detected an attempt to connect to an object which has already been destroyed in {WhereEverYourFileIs.php}

Assumption: Thread dies while running due to assignment of array. This behaviour is wether catched nor documeneted or obvisious. In addition, there is no workaround for this use case given.

krakjoe:
$this->result = ['blahkeks', $this->getThreadId()];

This sets ```$this->result``` to a ```Volatile``` object, the ```Volatile``` object belongs to the ```Thread```. When the ```Thread``` is destroyed, so is the ```Volatile``` object, and when the main context attempts to access it, you will raise an error.

If you really meant to use an array, explicitly cast:

    $this->result = (array) ['blahkeks', $this->getThreadId()];

This is expected behaviour.

webdes87:
Ok, I tried it and when (array)-casting it, the php.exe does not seem to crash. Thanks for that hint :)

But can you please enlighten my eyes, why the cast from [] to (array) [] is neccesary to not crash php.exe? And what this is related to volatile, since to my knowledge (Java) it just prevents some internal caching or other optimizations that might result in no write operation?

Under which circumstances can such crashes also happen?

[...]

PR #668, opened 2017-01-25
Enhance Pool example to show garbage collector getting array

paulobezerr:
This is a complement to be more easy to see the needed of casting the
``\Thread`` class variable to array to be possible recover by collector.

@krakjoe, sorry for posting some error in another user's thread. I discovered my problem and here I put some contribution to Pool example.
This was possible because of this issue: https://github.com/krakjoe/pthreads/issues/610.
Thank you!

### DIFF
--- a/examples/Pooling.php
+++ b/examples/Pooling.php
@@ -45,7 +45,35 @@ class WebWork extends Threaded {
 			$logger->log("%s has PDO in Thread #%lu", 
 				__CLASS__, $this->worker->getThreadId());
 		}
+
+		$this->setData(array(
+			'class' => __CLASS__,
+			'threadId' => $this->worker->getThreadId()
+		));
 	}
+
+	/**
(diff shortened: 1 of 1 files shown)


paulobezerr:
I forgot to check if the garbage variable is complete.
It is updated now.
Thank you!

tpunt:
I don't think these additions are something we want to show. It relies on the garbage collector to fetch data, which simply isn't a reliable thing to do, since its behaviour is not deterministic.

You also mentioned that it is not possible to fetch the (array-based) data if it has been converted into a `Volatile` object (which is also a `Threaded` object) - this isn't really true. `Threaded` objects are tied to the context in which they are created. This means that if data is going to be stored inside of a `Volatile` object (from the implicit coercion of an array, for example), then that object should be created in the outer-most scope in which it will be used (in this case, the main thread).

paulobezerr:
ok, thanks for explanation, @tpunt.

Issue #674, opened 2017-02-08
mysqli example isn't working


c9s:
Environment
==========

  * PHP: 7.1.0
  * pthreads: master branch
  * OS: OS X

Summary
======

Running the mysqli worker example returns exit code 0, however the result was not dumped.

It seems that `collect` method was not called:

```
$pool->collect(function($query) {
    die('test');
    printf($done = $query->getResult());
    return count($done);
});
```

Reproducing Code
==============

```
$pool->collect(function($query) {
    die('test');
    printf($done = $query->getResult());
    return count($done);
});
```

Expected Output
=============

    test

Actual Output
============

empty result.

c9s:
Here is the working version:

```php
/**
* This example illustrates best practice with regard to using MySQLi in multiple threads
*
* For convenience and simplicity it uses a Pool.
**/
class Connect extends Worker {
    public function __construct($hostname, $username, $password, $database, $port = 3306)
    {
        $this->hostname = $hostname;
        $this->username = $username;
        $this->password = $password;
        $this->database = $database;
        $this->port     = $port;
    }
    
    public function getConnection() {
        if (!self::$link) {
            self::$link = new mysqli(
                $this->hostname, 
                $this->username, 
                $this->password, 
                $this->database, 
                $this->port);
        }
        
        /* do some exception/error stuff here maybe */       
        return self::$link;
    }
    
    protected $hostname;
    protected $username;
    protected $password;
    protected $database;
    protected $port;
    
    /**
    * Note that the link is stored statically, which for pthreads, means thread local
    **/
    protected static $link;
}

class QueryJob extends Threaded {

    protected $done;

    public function __construct($sql) {
        $this->sql = $sql;
    }
    
    public function run() {
        $mysqli = $this->worker->getConnection();
        
        $result = $mysqli->query($this->sql);
        
        if ($result) {    
            while (($row = $result->fetch_assoc())) {
                $rows[] = $row;
            }
        }
        $this->result = $rows;
        $this->done = true;
    }

    public function getDone() {
        return $this->done;
    }
    
    public function getResult() {
        return $this->result;
    }
    
    protected $sql;
    protected $result;
}

$pool = new Pool(4, "Connect", ["localhost", "root", "", "mysql"]);
$pool->submit(new QueryJob("SHOW PROCESSLIST;"));
$pool->submit(new QueryJob("SHOW PROCESSLIST;"));
$pool->submit(new QueryJob("SHOW PROCESSLIST;"));
[...]

PR #675, opened 2017-02-08
Add failing test case for #674

c9s:
This is the failing test case for #674

### DIFF
--- a/tests/workers-job-array.phpt
+++ b/tests/workers-job-array.phpt
@@ -1,0 +1,39 @@
+--TEST--
+Test pthreads workers with array property
+--DESCRIPTION--
+This test verifies that threaded object property can return array value.
+--FILE--
+<?php
+
+class Job extends Threaded {
+
+    public function run()
+    {
+        $this->result = [1,2,3,4];
+    }
+
+    public function getResult()
+    {
+        return $this->result;
+    }
+
+    public function isGarbage() : bool {
+        return count($this->result);
+    }
+    
+    public $result;
+}
+
+$worker = new Worker;
+$worker->start();
+$worker->stack($jobs[] = new Job());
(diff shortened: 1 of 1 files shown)


tpunt:
As Joe mentioned on the issue, the example is simply outdated. I noticed the following things wrong with the test snippet you gave:
 - You are creating the results array inside of the new thread. This array is implicitly casted to a `Volatile` object (a `Threaded` object), which means it is tied to the `Job` object. When this object has finished running, the result (`Volatile`) object will be destroyed, causing pthreads to throw a `RuntimeException` when trying to access it. To prevent this from happening, it should be created in the constructor (and therefore by the main thread).
 - There's no need to define the `isGarbage` method.
 - The worker object should not be `join`ed - in particular, you are attempting to collect its garbage and shut it down again afterwards.

I'd rewrite your example test as follows:
```php
<?php

class Job extends Threaded
{
    public $result;

    public function __construct()
    {
        $this->result = new Volatile(); // create this in the outer most thread that it is needed
    }

    public function run()
    {
        $this->result[] = 1;
        $this->result[] = 2;
        $this->result[] = 3;
        $this->result[] = 4;
    }

    public function getResult()
    {
        return $this->result;
    }
}

$worker = new Worker;
$worker->start();
$jobs = [];

for ($i = 0; $i < 4; ++$i) {
    $worker->stack($jobs[] = new Job());
}

while ($worker->collect()); // collect the garbage to free memory used

$worker->shutdown(); // shutdown the worker to make sure it has completely finished before attempting to access job results

foreach ($jobs as $job) {
    print_r($job->getResult());
}
```

Given that this is being tracked in your other issue, I'm closing this PR.

c9s:
Thank you

Issue #666, opened 2017-01-23
__PHP_Incomplete_Class + Segfault


SirSnyder:
Hi Joe,

my application loads some classes at first. With a call of Thread::start(), many warnings like "Warning: Class __PHP_Incomplete_Class has no unserializer in Unknown on line 0" appear in the console.

[coredump.txt](https://github.com/krakjoe/pthreads/files/724105/coredump.txt)

Example code will follow.

Ubuntu with php 7.1.2-dev and pthreads latest master. 

 Thx

SirSnyder:
Segfault example code

```
function objectHash($object)
    {
        return crc32(spl_object_hash($object));
    }

    trait ClassInfo
    {
        /** @var int */
        private $intObjectHash;
    }

    class ThreadedMap extends Threaded
    {
        use ClassInfo;
    }

    class LiquidObject extends Threaded
    {
        use ClassInfo;
    }

    class ManagedInstanceProvider extends Threaded
    {
        use ClassInfo;

        const FREE = 0;
        const USED = 1;

        /** @var \Threaded */
        private $objIndex;

        /** @var \Threaded */
        private $objStack;

        /** @var \Threaded */
        private $objHashMap;

        /** @var bool */
        private $blnActive;

        /** @var string */
        private $strClassname;

        /** @var int */
        private $intLastIndex;

        /** @var int */
        private $intNumFreeIndizes;


        public function __construct(?string $classname)
        {
            $this->strClassname         = $classname;
            $this->intLastIndex         = 0;
            $this->intNumFreeIndizes    = 0;
            $this->blnActive            = false;
        }


        public function manage():void
        {
            $this->objStack->synchronized(function ()
            {
                if($this->intNumFreeIndizes === 0 && $this->objStack->count() === $this->intLastIndex)
                {
                    $objInstance = $this->createInstance();

                    $this->objStack[] = $objInstance;

                    $intPosition = $this->objStack->count() - 1;

                    $this->objIndex[$intPosition] = self::FREE;

                    $this->objHashMap[objectHash($objInstance)] = $intPosition;

                    $this->objStack->notify();
                }
            });
        }


        public function get()
        {
            return $this->objStack->synchronized(function ()
            {
                if(!$this->isActive())
                {
                    $this->activate();
                }
                $objInstance = null;

                while($objInstance === null)
                {
                    if($this->intNumFreeIndizes > 0)
                    {
                        $objInstance = $this->findFreeInstance();
                        continue;
                    }

[...]

PR #683, opened 2017-03-02
#666 fix Warning: Class __PHP_Incomplete_Class has no unserializer in…

SirSnyder:
… Unknown on line 0

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -48,7 +48,27 @@ static  zend_trait_method_reference * pthreads_preparation_copy_trait_method_ref
 static void pthreads_prepared_resource_dtor(zval *zv); /* }}} */
 
 /* {{{ */
-static zend_class_entry* pthreads_copy_entry(pthreads_object_t* thread, zend_class_entry *candidate) {
+static void pthreads_prepared_entry_static_members(pthreads_object_t* thread, zend_class_entry *candidate, zend_class_entry *prepared) {
+	if (candidate->default_static_members_count) {
+		int i;
+		prepared->default_static_members_table = (zval*) ecalloc(
+			sizeof(zval), candidate->default_static_members_count);
+		prepared->default_static_members_count = candidate->default_static_members_count;
+		memcpy(prepared->default_static_members_table,
+		       candidate->default_static_members_table,
+			sizeof(zval) * candidate->default_static_members_count);
+
+		for (i=0; i<prepared->default_static_members_count; i++) {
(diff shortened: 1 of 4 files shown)


src/prepare.c
```diff
@@ -576,13 +588,20 @@ static inline void pthreads_prepare_closures(pthreads_object_t *thread) {
...
+			pthreads_prepared_entry_internal(thread, entry, 0);
+		}
+	} ZEND_HASH_FOREACH_END();
+
+	ZEND_HASH_FOREACH_STR_KEY_PTR(PTHREADS_CG(thread->creator.ls, class_table), name, entry) {
+		if ((prepared = zend_hash_find_ptr(PTHREADS_CG(thread->local.ls, class_table), name)) && prepared->type != ZEND_INTERNAL_CLASS) {
+			pthreads_prepared_entry_static_members(thread, entry, prepared);
 		}
-	} ZEND_HASH_FOREACH_END(); 
+	} ZEND_HASH_FOREACH_END();
```
tpunt: This is a bit of a nitpick, but if you inverse the loops, then you can skip a check on each iteration:
```c
ZEND_HASH_FOREACH_STR_KEY_PTR(PTHREADS_CG(thread->local.ls, class_table), name, entry) {
    if (entry->type != ZEND_INTERNAL_CLASS) {
        pthreads_prepared_entry_static_members(
            thread,
            zend_hash_find_ptr(PTHREADS_CG(thread->creator.ls, class_table), name),
            entry
        );
    }
} ZEND_HASH_FOREACH_END();
```

I cannot immediately see anything wrong with this, since a child thread will only ever have classes that exist in the parent thread.

tpunt:
This doesn't cover all cases, it seems. For example, the following errors:
```php
class SecondClass extends FirstClass {}
class FirstClass extends \Threaded
{
	public static $prop = [];
}
FirstClass::$prop[] = new SecondClass();
$thread = new class extends Thread {
    public function run() {
        var_dump(FirstClass::$prop);
    }
};
$thread->start() && $thread->join();
```

This can easily be changed by not copying the statics when preparing parent classes in `pthreads_copy_entry`:
```c
prepared->parent = pthreads_prepared_entry_internal(thread, candidate->parent, 0);
```

There could be other edge cases around the preparing of classes, but I haven't had the time to check for them.

SirSnyder:
@tpunt both fixed and I've added your example as test.

tpunt:
Merged, thanks!

PR #684, opened 2017-03-02
fix usources size

SirSnyder:


pp3345:
Wouldn't it be better to use macros for this instead of a do-while loop?

tpunt:
Merged in a83eca3. Thanks!

PR #685, opened 2017-03-02
#666 - don't override creator TSRMLS

SirSnyder:


src/object.c
```diff
@@ -253,9 +253,11 @@ int pthreads_connect(pthreads_object_t* source, pthreads_object_t* destination)
 
 			pthreads_monitor_free(destination->monitor);
 		}
+		pthreads_ident_t creator = destination->creator;
```
krakjoe: we use C89 decl, reposition this to top of block please

SirSnyder:
Declaration is now on top of block, ok?

src/object.c
```diff
@@ -239,7 +239,9 @@ void pthreads_current_thread(zval *return_value) {
 
 /* {{{ */
 int pthreads_connect(pthreads_object_t* source, pthreads_object_t* destination) {
-	if (source && destination) {
+	pthreads_ident_t destCreator = destination->creator;
```
tpunt: Can you place this declaration inside of the `if` statement below, please. The below condition is checking to make sure `destination` is not `NULL`, so there must be some case when it can be `NULL` (in which case, the current code will segfault by trying to deref a `NULL` pointer (`destination`)).

tpunt:
Other than my single comment above, this looks good to me.

SirSnyder:
@tpunt oh, yes of course. updated

tpunt:
Merged, thanks!

Issue #659, opened 2017-01-15
Abort Trap Error on Assigning Anonymous Classes to Properties


tpunt:
Environment
==========

  * PHP: PHP 7.1.0 (cli) ( ZTS DEBUG )
  * pthreads: 3.1.7dev
  * OS: Mac

Reproducing Code
==============

```php
$task = new class extends Thread {
    public function run()
    {
        $this->prop = new class {};
    }
};

$task->start() && $task->join();
```

krakjoe:
I had a go at #658 and #659 this morning ... but failed ... I'll have to come back to it ...

tpunt:
Resolved by #686.

Issue #658, opened 2017-01-15
Segfault on Anonymous Class with a (not yet started) Worker


tpunt:
Environment
==========

  * PHP: PHP 7.1.0 (cli) ( ZTS DEBUG )
  * pthreads: 3.1.7dev
  * OS: Mac

Reproducing Code
==============

```php
$worker = new Worker();
$task = new class extends Threaded {
    public static $a;
};
$task::$a = new class extends Threaded {};
$worker->start();
$worker->stack($task);
$worker->shutdown();
```

The worker must not be started before the static property assignment. `Pool`s are also affected:

```php
$task = new class extends Threaded {
    public static $a;
};
$task::$a = new class extends Threaded {};
$pool = new Pool(2);
$pool->submit($task);
$pool->shutdown();
```

krakjoe:
I had a go at #658 and #659 this morning ... but failed ... I'll have to come back to it ...

tpunt:
Resolved by #686.

PR #686, opened 2017-03-05
Feature/anonymous class support

### LINKED ISSUE #505: Anonymous class crashes Thread
Hi @krakjoe,
a new instance of an anonymous class within a thread context crashes the application. A simple example:

```
class Test extends Thread
    {
        public function run()
        {
            new class {};
        }
    }

    $objThread = new Test();
    $objThread->start();
    $objThread->join();
```

Thx

---

Win7 x64
pthreads 3.0.8 (http://windows.php.net/downloads/pecl/releases/pthreads/3.0.8/ is not ready!)
php7 rc 5 (master / build by Jan-E)

SirSnyder:
anonymous class support, fix #505 #658 #659

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -51,6 +51,10 @@ static void pthreads_prepared_resource_dtor(zval *zv); /* }}} */
 static void pthreads_prepared_entry_static_members(pthreads_object_t* thread, zend_class_entry *candidate, zend_class_entry *prepared) {
 	if (candidate->default_static_members_count) {
 		int i;
+
+                if(prepared->default_static_members_table != NULL) {
+                        efree(prepared->default_static_members_table);
+                }
 		prepared->default_static_members_table = (zval*) ecalloc(
 			sizeof(zval), candidate->default_static_members_count);
 		prepared->default_static_members_count = candidate->default_static_members_count;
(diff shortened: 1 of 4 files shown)


tpunt:
This PR shouldn't have commits from other PRs in it - it just makes it a lot harder to review.

SirSnyder:
Other PRs removed

src/prepare.c
```diff
@@ -256,6 +240,10 @@ while(0)
 	
 	if (candidate->default_properties_count) {
 		int i;
+		
+		if(prepared->default_properties_table != NULL) {
```
krakjoe: I'm not sure how this can happen ?

src/prepare.c
```diff
@@ -276,6 +264,9 @@ while(0)
 
 	if (candidate->default_static_members_count) {
 		int i;
+		if(prepared->default_static_members_table != NULL) {
```
krakjoe: As above, not sure how that happens ?

src/prepare.c
SirSnyder: zend_class_entry of anonymous classes already exists per interpreter instance at thread creation, like internal classes. But they are incomplete, especially the default_properties_table and default_static_members_table. I guess the anon class zend_class_entry gets "constructed" at compile time. But final property and static member info is not known at compile time, first at runtime.

tpunt:
@SirSnyder Can you rebase this please? I hope to review it again sometime this weekend.

SirSnyder:
@tpunt rebased.

krakjoe:
@tpunt I guess you reviewed this already, or shall I do it ?

tests/no-inheritance-anon-class.phpt
```diff
@@ -0,0 +1,16 @@
...
+Test fix for #659
+--DESCRIPTION--
+Unbound anon class causing segfaults, we delay copy but still cannot serialize the anon
+--FILE--
+<?php
+$task = new class extends Thread {
+    public function run()
+    {
+        $this->prop = new class {};
+		var_dump($this->prop); /* we do expect null: anon classes cannot be serialized */
```
dktapps: This feels strange... wouldn't it be better to have it throw an exception here? suddenly producing null is probably not going to be expected behaviour.

tpunt:
@krakjoe I gave this a quick review a while ago, and it seemed ok then. I won't be able to give it a thorough review until sometime next week, however.

tpunt:
@krakjoe I gave this a quick review a while ago, and it seemed ok then. I won't be able to give it a thorough review until sometime next week, however.

tests/no-inheritance-anon-class.phpt
tpunt: I'd say this behaviour is entirely expected. When an object of an unserialisable class is assigned to a property of a `Threaded` object, it will assign `null` instead. This can already be seen with the `SimpleXMLElement`, `HashContext`, and `SplFileInfo` classes, for example. 

Whether these semantics should be changed (to throw an exception, for example) is orthogonal to this PR.

tpunt:
@krakjoe I've given this patch another review and it seems fine. I'll leave this open, though, in case you would like to give it a review first.

Also, with this patch, the thread preparation logic becomes rather convoluted. If this gets merged, then I may have a look to see how things can be refactored to make the code easier to reason about (or at the very least, add a few comments about why certain checks are necessary).

SirSnyder:
Good to read. With this patch finally merged, I could submit the next bug fix patch ;-)

tpunt:
Joe seems quite busy at the moment, so I've merged this myself. Thanks!

SirSnyder:
Thanks @tpunt!

Issue #681, opened 2017-02-22
Threaded Memory leak


tmtbe:
How to destroy Threaded？
```php
<?php
ini_set('zend.enable_gc', true);
class Work extends Threaded {
    public function __construct()
    {
    }

    public function run(){
        print_r(1);
    }
}

$my = new Worker();
$work = new Work($my);
$my->start();
$my->stack($work);
cli_set_process_title('swdd');
while (true){
    $my->collect(function (Threaded $t)use($my){
        $work = new Work();
        $my->stack($work);
        return true;
    });
}
```

The above code is running, the memory will continue to rise.

Jackysi:
I've had same similar issue for months, too bad I have zero knowledge about C to fix this. I was hoping krakjoe has more time to fix this but sadly the author is just soo busy ^^. I can understand, but this extension is so useful for those that want to use PHP for simple daemons and have no knowledge or will to learn C/C++/C#

SirSnyder:
@tmtbe @Jackysi should be fixed, see my patch

Jackysi:
so this did fix the issue, I can't test right now cause I kinda removed the threading code?

saniel:
Hi guys, I've noticed the memory leak too. I'm using PHP 7.1 with pthreads from master branch cloned on april 27th.
I'm using following code:
```
<?php
$worker = new \Worker();
$worker->start();

sleep(3); // to get time to manually get memory usage

$i = 500000;
while ($i) {
	$i--;
	$work = new \Threaded();
	$worker->stack($work);
	
//	while ($worker->getStacked()) {};
//	$worker->collect();
}

while ($worker->getStacked()) {};
$worker->collect();

$worker->shutdown();

echo "done\n";
while (1) sleep(1);
```

I get memory usage from "ps aux".

_Memory used just after starting new worker thread:_
daniel   23561  1.3  0.3 **287456 28772** pts/1    Sl+  12:07   0:00 php test.php

_Memory use when executed as it is:_
daniel   23354 72.6  5.5 **683352 448592** pts/1   S+   11:55   0:05 php test.php

_When I uncomment the commented code, i.e. wait for previous work to be done before stacking next I get significantly less memory used, but still growing indefinitely, when looping forever._
daniel   23114 10.4  0.9 **284880 75700** pts/1    S+   11:54   0:05 php test.php

When I create my own work class which stores some initial data in it's constuctor, the memory  is consumed more.

I hope this helps a bit to identify source of the issue.

[...]

PR #688, opened 2017-03-08
fix #580 #681

### LINKED ISSUE #580: Some internal data structures seem to stick around and start filling memory
# Environment
- PHP: PHP 7.0.4 (cli) (built: Mar 12 2016 14:07:47) ( ZTS )*
- pthreads: Version => 3.1.7dev
- OS: Ubuntu 14.04 LTS (HVM) (3.13.0-79-generic)
# Summary

When using Volatile arrays, it seems that 'unsetting' the array elements doesnt fully remove the data. The thread that unsets them seems fine, var_dump($this) & var_export($this,true) shows that. But the thread that actually set the data still has the data locally somewhere. It doesnt show with var_dump($this), but it does  show with var_export($this,true). Not sure why, and I couldn't find what the diff was with dump or export.

If you loop this to make the creator thread create batches of 100000 and then the deleting thread delete them all, and then loop, and leave it running, eventually all memory is consumed.

It seems to be to do with myRec being 'Threaded'. If its not threaded its fine, but I then cant set individual members like in setData1() as it just ignores the set.

Sorry for a not so simple test case, but this was the simplest way I could show it happening in a test case.

(Ignore the unsafe waiting and notifying, its just for the example, and the sleep is just to make it easier to show this happening rather than a more complex wait/notify.)
# Reproducing Code

```
<?php

class myRec extends Threaded {
    public $data1;
    public $data2;
}

class storage extends Threaded {

        private $store = array();

        public function addRecord($key,$data) {
                $this->store[$key] = $data;
        }

        public function delAll() {
                foreach ($this->store as $key => $data) {
                        unset($this->store[$key]);
                }
        }

        public function getData1($key) {
                return $this->store[$key]->data1;
        }
        public function setData1($key,$data) {
                $this->store[$key]->data1 = $data;
        }
}

class myThread extends Thread
{

    private static $type;

    public function __construct($type,$storage)
    {
        self::$type = $type;
        $this->storage = $storage;
    }

    public function run()
    {

        if (self::$type === 1) {
                for ($i=0;$i<5;$i++) {
                        $r = new  myRec();
                        $r->data1 = 'data'.$i;
                        $this->storage->addRecord($i,$r);
                }

                $this->storage->setData1(1,'newData');
                var_dump($this->storage->getData1(1));

                $this->synchronized(function() {$this->wait();});

                echo "CREATING THREADS THIS:\n";
                var_dump($this);

                $thisExport = explode("\n",var_export($this,true));
                foreach ($thisExport as $data) {
                    print 'CREATING THREAD EXPORT|'.$data . "\n";
                }


        }

        if (self::$type === 2) {
                $this->storage->delAll();

                $thisExport = explode("\n",var_export($this,true));
                foreach ($thisExport as $data) {
                    print 'DELETING THREAD EXPORT|'.$data . "\n";
                }


        }
    }
}

$storage = new storage();

$threads[0] = new myThread(1,$storage);
$threads[0]->start();

sleep(3);// Unsafe, but just for this example to let thread 1 add the 5 recs

$threads[1] = new myThread(2,$storage);
$threads[1]->start();
$threads[1]->join();

// Now thread 1 should not have any data
$threads[0]->synchronized(function() use($threads) {$threads[0]->notify();});

?>
```
# Expected Output

```
string(7) "newData"
DELETING THREAD EXPORT|myThread::__set_state(array(
DELETING THREAD EXPORT|   'storage' =>
DELETING THREAD EXPORT|  storage::__set_state(array(
DELETING THREAD EXPORT|     'store' =>
DELETING THREAD EXPORT|    Volatile::__set_state(array(
DELETING THREAD EXPORT|    )),
DELETING THREAD EXPORT|  )),
DELETING THREAD EXPORT|))
CREATING THREADS THIS:
object(myThread)#1 (1) {
  ["storage"]=>
  object(storage)#5 (1) {
    ["store"]=>
    object(Volatile)#6 (0) {
    }
  }
}
CREATING THREAD EXPORT|myThread::__set_state(array(
CREATING THREAD EXPORT|   'storage' =>
CREATING THREAD EXPORT|  storage::__set_state(array(
CREATING THREAD EXPORT|     'store' =>
CREATING THREAD EXPORT|    Volatile::__set_state(array(
CREATING THREAD EXPORT|    )),
CREATING THREAD EXPORT|  )),
CREATING THREAD EXPORT|))
```
# Actual Output

```
string(7) "newData"
DELETING THREAD EXPORT|myThread::__set_state(array(
DELETING THREAD EXPORT|   'storage' =>
DELETING THREAD EXPORT|  storage::__set_state(array(
DELETING THREAD EXPORT|     'store' =>
DELETING THREAD EXPORT|    Volatile::__set_state(array(
DELETING THREAD EXPORT|    )),
DELETING THREAD EXPORT|  )),
DELETING THREAD EXPORT|))
CREATING THREADS THIS:
object(myThread)#1 (1) {
  ["storage"]=>
  object(storage)#9 (1) {
    ["store"]=>
    object(Volatile)#10 (0) {
    }
  }
}
CREATING THREAD EXPORT|myThread::__set_state(array(
CREATING THREAD EXPORT|   'storage' =>
CREATING THREAD EXPORT|  storage::__set_state(array(
CREATING THREAD EXPORT|     'store' =>
CREATING THREAD EXPORT|    Volatile::__set_state(array(
CREATING THREAD EXPORT|       0 =>
CREATING THREAD EXPORT|      myRec::__set_state(array(
CREATING THREAD EXPORT|         'data1' => 'data0',
CREATING THREAD EXPORT|         'data2' => NULL,
CREATING THREAD EXPORT|      )),
CREATING THREAD EXPORT|       1 =>
CREATING THREAD EXPORT|      myRec::__set_state(array(
CREATING THREAD EXPORT|         'data1' => 'newData',
CREATING THREAD EXPORT|         'data2' => NULL,
CREATING THREAD EXPORT|      )),
CREATING THREAD EXPORT|       2 =>
CREATING THREAD EXPORT|      myRec::__set_state(array(
CREATING THREAD EXPORT|         'data1' => 'data2',
CREATING THREAD EXPORT|         'data2' => NULL,
CREATING THREAD EXPORT|      )),
CREATING THREAD EXPORT|       3 =>
CREATING THREAD EXPORT|      myRec::__set_state(array(
CREATING THREAD EXPORT|         'data1' => 'data3',
CREATING THREAD EXPORT|         'data2' => NULL,
CREATING THREAD EXPORT|      )),
CREATING THREAD EXPORT|       4 =>
CREATING THREAD EXPORT|      myRec::__set_state(array(
CREATING THREAD EXPORT|         'data1' => 'data4',
CREATING THREAD EXPORT|         'data2' => NULL,
CREATING THREAD EXPORT|      )),
CREATING THREAD EXPORT|    )),
CREATING THREAD EXPORT|  )),
CREATING THREAD EXPORT|))
```

SirSnyder:
Cleanup on synchronized or reading a property is inconsistent, therefore cleanup should also happen on writing a property. Moreover the cleanup behavior on synchronized is misleading. 

```
class myRec extends Threaded {
	public $data1;
	public $data2;
}

class storage extends Threaded {

	private $store = array();

	public function addRecord($key,$data) {
        	$this->store->synchronized(function () use ($key, $data)
        	{
			$this->store[$key] = $data;
		});
	}

	public function delAll() {
        	$this->store->synchronized(function ()
		{
			foreach($this->store as $key => $data)
			{
				unset($this->store[$key]);
			}
        	});
	}

	public function countRecs() {
		return count($this->store);
	}
}

class myThread extends Thread
{

	private static $type;

	public function __construct($type,$storage)
	{
		self::$type = $type;
		$this->storage = $storage;
	}

	public function run()
	{

		if (self::$type === 1) {
			while (true) {

				//if($this->storage->countRecs() == 0)
				{
					@$loop++;
					for($i = 0; $i < 30000; $i++)
					{
						$r = new  myRec();
						$r->data1 = 'data' . $i;
						$this->storage->addRecord($loop . '|' . $i, $r);
					}
				}

				sleep(5);

			}
		}

		if (self::$type === 2) {
			while (true) {
				//if($this->storage->countRecs() >= 30000)
				{
					$this->storage->delAll();
				}

				sleep(1);
				print "RECS:".$this->storage->countRecs()."\n";
			}

		}
	}
}

ini_set('memory_limit', '-1');

$storage = new storage();

$threads[0] = new myThread(1,$storage);
$threads[0]->start();

$threads[1] = new myThread(2,$storage);
$threads[1]->start();
$threads[1]->join();
```
This is a fix for #681 as well.

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -336,7 +336,8 @@ int pthreads_store_write(zval *object, zval *key, zval *write) {
 
 				if (zend_hash_update_ptr(threaded->store.props, keyed, storage)) {
 					result = SUCCESS;
-				} else zend_string_release(keyed);
+				}
+				zend_string_release(keyed);
 			}
 		}
 		pthreads_monitor_unlock(threaded->monitor);
@@ -351,6 +352,7 @@ int pthreads_store_write(zval *object, zval *key, zval *write) {
 				normal refcounting, we'll just never read the reference
 			*/
 			rebuild_object_properties(&threaded->std);
+
 			if (Z_TYPE(member) == IS_LONG) {
 				zend_hash_index_update(threaded->std.properties, Z_LVAL(member), write);
 			} else {
@@ -358,7 +360,8 @@ int pthreads_store_write(zval *object, zval *key, zval *write) {
 				if (zend_hash_update(
 					threaded->std.properties, keyed, write)) {
 					result = SUCCESS;
-				} else zend_string_release(keyed);
+				}
+				zend_string_release(keyed);
 			}
 			Z_ADDREF_P(write);
 		}


src/store.c
```diff
@@ -351,14 +352,18 @@ int pthreads_store_write(zval *object, zval *key, zval *write) {
 				normal refcounting, we'll just never read the reference
 			*/
 			rebuild_object_properties(&threaded->std);
+			
+			pthreads_store_sync(object);
```
tpunt: It's not immediately clear to me why this is needed - is it related to the memory leaks you're fixing? And do you have a test case for this?

SirSnyder: A test case might be difficult. I ran the script above for a certain time and watched the memory consumption. Any idea to trace this in a test script? 

I've added pthreads_store_sync at this point to guarantee consistency with the behavior of reading properties.

tpunt: Testing the memory leak isn't really necessary, since it's quite unlikely a regression will be introduced here. I just wasn't sure how syncing the pthreads store linked to the memory leaks (it doesn't, but I see that you're using this PR to target different bugs).

krakjoe: This has the potential to make unsynchronized writes extremely slow ... I don't like it ... would be good to find another way to do whatever you are trying to do ...

SirSnyder: Producer/Consumer: Producer Thread A writes properties and never reads them. Consumer Thread B reads/removes properties. Table std.properties in Thread A fills up, endlessly. I could limit `pthreads_store_sync` to Volatile objects, that might be a solution you like?

SirSnyder: I have again tested the script, without the pthreads_store_sync call. I must admit the memory isn't growing. My fault, I must have done something wrong.

tpunt:
@SirSnyder I'm also having trouble replicating the issue you're receiving without the syncing of the pthreads store. Do you have a smaller reproducing script that explicitly shows a difference with and without the syncing of stores on property writes?

SirSnyder:
@tpunt that must have been an erroneous observation in htop. My fault. I've removed the pthreads_store_sync call.

tpunt:
Merged in 1dfeea7. Thanks!

PR #701, opened 2017-05-03
Bug/return type hanging thread

SirSnyder:
Wrong logical operator, should be equal to the other branch. A fake typ in ```instanceof_function``` leads to a infinite loop.

### DIFF
--- a/php_pthreads.c
+++ b/php_pthreads.c
@@ -156,11 +156,11 @@ static inline void pthreads_execute_ex(zend_execute_data *data) {
 /* {{{ */
 static inline zend_bool pthreads_verify_type(zend_execute_data *execute_data, zval *var, zend_arg_info *info) {
 	pthreads_object_t *threaded = NULL;
-
+        
 	if (!var ||
 #if PHP_VERSION_ID >= 70200
-		!ZEND_SAME_FAKE_TYPE(ZEND_TYPE_CODE(info->type), Z_TYPE_P(var)) &&
-		!ZEND_TYPE_IS_CLASS(info->type) ||
+                Z_ISNULL_P(var) ||
+                !ZEND_TYPE_IS_CLASS(info->type) ||
 #else
 		!ZEND_SAME_FAKE_TYPE(info->type_hint, Z_TYPE_P(var)) ||
 		info->type_hint != IS_OBJECT ||
--- a/tests/verify-overload-71.phpt
+++ b/tests/verify-overload-71.phpt
@@ -1,0 +1,31 @@
+--TEST--
+Test VERIFY_RETURN_TYPE overload - PHP 71 compliant
+--DESCRIPTION--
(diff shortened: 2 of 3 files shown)


SirSnyder:
hm, some tests failed. I'll look at it

SirSnyder:
Existing test extended, all tests passed.

php_pthreads.c
```diff
@@ -156,11 +156,11 @@ static inline void pthreads_execute_ex(zend_execute_data *data) {
 /* {{{ */
 static inline zend_bool pthreads_verify_type(zend_execute_data *execute_data, zval *var, zend_arg_info *info) {
 	pthreads_object_t *threaded = NULL;
-
+        
 	if (!var ||
 #if PHP_VERSION_ID >= 70200
-		!ZEND_SAME_FAKE_TYPE(ZEND_TYPE_CODE(info->type), Z_TYPE_P(var)) &&
-		!ZEND_TYPE_IS_CLASS(info->type) ||
+                Z_ISNULL_P(var) ||
```
krakjoe: feels like null check is wrong ...

it may break nullable types ... may be better to introduce null check with allow_null verification from arg info ...

SirSnyder: The null check is necessary for nullable types. In case a nulled $var gets returned `return ($var = null);`, Z_ISNULL_P(var) is true. This is covered by tests/verify-overload-71.phpt. A simple ```return null``` doesn't trigger `pthreads_verify_type`.

tpunt: Would it not be enough to simply change the `&&` into an `||`? The current logic of:
```
!ZEND_SAME_FAKE_TYPE(ZEND_TYPE_CODE(info->type), Z_TYPE_P(var)) &&
!ZEND_TYPE_IS_CLASS(info->type)
```

doesn't work, because if it is not the same (fake) type, then the anything-but-a-class check will always fail. If the type is not a class, then we want to (and must) short circuit, since a class type is presumed when the `instanceof_function` function is used for the next check. Or am I missing something?

SirSnyder: A change from `&&` to `||` at that point would work for primitive types(fake type?) only, but not for classes. In case of an object/class type, `ZEND_SAME_FAKE_TYPE` would let the condition match and falsely return. In generally the `ZEND_SAME_FAKE_TYPE` check is unnecessary in `pthreads_verify_type`, because `pthreads_verify_type` is only called when the types doesn't match. If the type is a class, do `instanceof_function`. In every other case, type is a primitive or a null var, return immediately and let the script fail.

SirSnyder: @krakjoe still feedback required?

krakjoe: there are merge conflicts now, this code was refactored ... don't know if bug is still present, please test ...

krakjoe:
As far as I can tell, this was fixed when we prepared for 7.2 ... sorry it took so long :)

PR #702, opened 2017-05-15
Update stub

dktapps:
Threaded::wait() parameter is optional


tpunt:
Merged in 1997397. Thanks!

PR #712, opened 2017-06-05
Resolve bug #711 (class declaration after new thread creation)

### LINKED ISSUE #711: Worker segfault when stacking a task defined after thread start
Environment
==========

  * PHP: 7.0.19 ZTS
  * pthreads: 3.1.6
  * OS: Windows 10 x64

Summary
======
Defining the class for a Worker job causes a segfault when stacked if the class was defined after the worker was started.

Also observed similar behaviour on PHP 7.1 with the latest master on Linux.

Starting the Worker thread with PTHREADS_INHERIT_ALL also resolves the issue, however I find a segmentation fault unexpected.

Reproducing Code
==============
Working:
```
<?php

$worker = new \Worker();
$worker->start();

class task extends \Threaded implements \Collectable{
    
    public function __construct(){
        var_dump("constructor");
    }
    
    public function run(){

    }
}

var_dump("stacking");
$worker->stack(new task());
var_dump("stacked");
while($worker->collect());
$worker->shutdown();
var_dump("ok");
```

Broken:
```
<?php

class task extends \Threaded implements \Collectable{
    
    public function __construct(){
        var_dump("constructor");
    }
    
    public function run(){

    }
}

$worker = new \Worker();
$worker->start();

var_dump("stacking");
$worker->stack(new task());
var_dump("stacked");
while($worker->collect());
$worker->shutdown();
var_dump("ok");
```

Expected Output
=============
```
string(8) "stacking"
string(11) "constructor"
string(7) "stacked"
string(2) "ok"
```

Actual Output
============
```
string(8) "stacking"
string(11) "constructor"
string(7) "stacked"
```
followed by a segmentation fault.

tpunt:
This fixes #711.

I'm not sure on the full ramifications of this fix quite yet, though. The basic idea is to delay the copying of certain classes (on thread creation) until they are actually used (basically, lazy copying), since the class entry seems to be incomplete at the point of thread creation. Notably, the `create_object` member of `zend_class_entry` is `NULL`, and so the new `Threaded` object is not created by pthreads' constructor, but rather the built-in constructor. This causes issues down the line when attempting to fetch the pthreads object from the allocated Zend object.

It requires the class to be implementing an interface (though I haven't figured out why that is necessary, yet).

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -591,7 +591,7 @@ static inline void pthreads_prepare_classes(pthreads_object_t* thread) {
 	zend_string *name;
 	
 	ZEND_HASH_FOREACH_STR_KEY_PTR(PTHREADS_CG(thread->creator.ls, class_table), name, entry) {
-		if (!zend_hash_exists(PTHREADS_CG(thread->local.ls, class_table), name)) {
+		if (!zend_hash_exists(PTHREADS_CG(thread->local.ls, class_table), name) && ZSTR_VAL(name)[0] != '\0') {
 			pthreads_prepared_entry_internal(thread, entry, 0);
 		}
 	} ZEND_HASH_FOREACH_END();
--- a/tests/new-class-after-thread-creation.phpt
+++ b/tests/new-class-after-thread-creation.phpt
@@ -1,0 +1,18 @@
+--TEST--
+New class after thread creation
+--DESCRIPTION--
+Fixes bug #711. This requires a class to be declared after a new thread has been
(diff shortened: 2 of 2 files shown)


SirSnyder:
Your description looks similar to my anonymous class observations. The `create_object` member of anonymous `zend_class_entry` is also `NULL`. #686 would solve this issue too.

krakjoe:
Merged 082f2eff01ddfbbbda3d5223a331d58c40ec6489

Thanks.

Issue #733, opened 2017-08-06
Socket::accept() and Socket::setBlocking(false)


prolic:
I am not sure if this is the correct behaviour. But once `Socket::setBlocking(false)` is called, a call to `Socket::accept()` throws `RuntimeException: socket found in invalid state`.

When I remove that `setBlocking(false)` call, the thread hangs forever during `accept()` and waits for a connection. This leads to being impossible to shutdown the thread, as I cannot close the socket nor kill the thread (`Thread::kill()` has been removed).

Is this a bug? If not, how should I proceed?

Environment
==========

  * PHP: 7.2.0beta1
  * pthreads: latest master
  * OS: Ubuntu 16.04 LTS

SirSnyder:
I guess this is a bug. Should be easy to fix. Will give it a try the next days.

SirSnyder:
@tpunt please add the bug label

dktapps:
On a semi-related note, this is also a problem with resources... if you have an application with a thread waiting for a blocking stdin read to complete, for example, you can't stop the thread without user input, and are unable to kill it.

tpunt:
Reading through the [`accept`](http://man7.org/linux/man-pages/man2/accept.2.html) man page (since `Socket::accept` uses it internally) shows the following:
> If no pending connections are present on the queue, and the socket is
       not marked as nonblocking, accept() blocks the caller until a
       connection is present.  **If the socket is marked nonblocking and no
       pending connections are present on the queue, accept() fails with the
       error EAGAIN or EWOULDBLOCK.**

So it would seem that this is expected behaviour. I cannot offer much more advice on this at the moment, though, because I'm not well-versed in the sockets code (yet). I'll try to look more into this in the coming week.

prolic:
Btw: Will is the Tread::kill() method removed? With it I probably could ignore the setBlocking-problem and just kill the thread, when I want to stop the socket server.

tpunt:
I can think of two reasons for removing `Thread::kill`:
 1. If a thread acquires one of pthread's global mutexes, and you then kill it, the lock will not be released. This could probably be solved with Robust Mutexes, but it isn't something I've looked into.
 2. Killing a thread will not enable for that thread's memory to be released - it will be leaked until the main PHP process ends.

There could be other problems with it too - the above two points were just my guesses of why it was removed.

prolic:
Ok so I need a non blocking socket server, so I can stop him gracefully
then :-)

[...]

PR #734, opened 2017-08-06
add socket server example

prolic:
this example also includes pcntl usage. but there is an issue regarding on how to stop a thread, that is waiting for `Socket::accept()`, therefore stopping the socket server is not really good working, see: https://github.com/krakjoe/pthreads/issues/733

I hope we can get this one running :beers:


tpunt:
Given your recent comment in #733, I guess this PR has now been superseded by #743, so I'm closing this.

PR #740, opened 2017-08-23
extends verify-overload test

SirSnyder:
Just a small extension of an existing test


tpunt:
Just a minor nitpick, but pthreads uses tabs instead of spaces (which is why things are misaligned in the patch preview).

SirSnyder:
You're absolutely right, my netbeans expanded tabs to spaces. Changed.

tpunt:
Merged, thanks!

Issue #725, opened 2017-07-24
Function run time cache segmentation fault with PHP 7.1 & 7.2


dktapps:
Environment
==========

  * PHP: PHP 7.1.x and PHP 7.2.0dev
  * pthreads: 3.1.7dev
  * OS: Windows 10 x64 and Ubuntu 17.04 x64

Summary
======
When attempting to upgrade to PHP 7.1 (and 7.2) I've been experiencing a consistent segfault with function runtime caches being null when calling certain methods in worker tasks (in our project the segfault is occurring when [this](https://github.com/pmmp/PocketMine-MP/blob/php/7.2/src/pocketmine/scheduler/AsyncTask.php#L145) function is called). 

This issue did not occur on PHP 7.0 with the same pthreads source, leading me to believe a change to php-src is to blame for this bug.

Sadly I do not know how to reproduce this bug with minimal code, I've attempted to cut it down to a reproducing test case, but have been unable to produce one. 

Possible solutions
https://github.com/SirSnyder/pthreads/commit/fd8fac4675c20ef310c07e6c367bb599da7a6629
https://github.com/dktapps/pthreads/commit/56115b126d20f9cccc5bb9c9f26011888602cd94

Both of the above commits resolve the issue in my project, however I'm a noob here.

Sorry for such a useless issue, I hope someone is better able to identify the cause of this problem than I am.

Visual Studio backtrace:
```
>	php7ts.dll!ZEND_INIT_METHOD_CALL_SPEC_UNUSED_CONST_HANDLER(_zend_execute_data * execute_data) Line 28067	C
 	php7ts.dll!execute_ex(_zend_execute_data * ex) Line 59440	C
 	php7ts.dll!pthreads_execute_ex(_zend_execute_data * data) Line 150	C
 	php7ts.dll!ZEND_DO_FCALL_SPEC_RETVAL_USED_HANDLER(_zend_execute_data * execute_data) Line 1071	C
 	php7ts.dll!execute_ex(_zend_execute_data * ex) Line 59440	C
 	php7ts.dll!pthreads_execute_ex(_zend_execute_data * data) Line 150	C
 	php7ts.dll!ZEND_DO_FCALL_SPEC_RETVAL_UNUSED_HANDLER(_zend_execute_data * execute_data) Line 889	C
 	php7ts.dll!execute_ex(_zend_execute_data * ex) Line 59440	C
 	php7ts.dll!pthreads_execute_ex(_zend_execute_data * data) Line 150	C
[...]

PR #741, opened 2017-08-23
fix static property circular references during class preparation

SirSnyder:
I have no further details or tests for this bug fix, sorry.

### DIFF
--- a/src/pthreads.h
+++ b/src/pthreads.h
@@ -122,6 +122,7 @@ ZEND_BEGIN_MODULE_GLOBALS(pthreads)
 	zval  this;
 	HashTable resolve;
 	HashTable *resources;
+	zend_bool prepare_static_members;
 ZEND_END_MODULE_GLOBALS(pthreads)
 #	define PTHREADS_ZG(v) TSRMG(pthreads_globals_id, zend_pthreads_globals *, v)
 #   define PTHREADS_PID() PTHREADS_ZG(pid) ? PTHREADS_ZG(pid) : (PTHREADS_ZG(pid)=getpid())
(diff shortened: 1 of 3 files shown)


src/copy.h
```diff
@@ -288,7 +288,7 @@ static inline zend_function* pthreads_copy_internal_function(zend_function *func
 static zend_function* pthreads_copy_function(zend_function *function) {
 	zend_function *copy = zend_hash_index_find_ptr(&PTHREADS_ZG(resolve), (zend_ulong)function);
 	
-	if (copy) {
+	if (copy && copy->op_array.run_time_cache == NULL) {
```
tpunt: This change does not look related and it also looks testable. If this is the case, can you separate this out into a separate PR (with a test) to ensure that it is definitely applied?

tpunt:
@SirSnyder I've submitted a PR to refactor the class preparation logic (see #755). Given that this changes a fair amount of code you've written, could you please have a look over it to ensure that I haven't missed anything please? Thanks.

src/copy.h
dktapps: I think the bug that this change fixes was already resolved by the fix for #725. This was one of the patches I tested that resolved the bug.

tpunt: Oh right, thanks for the information. It looks like it is no longer possible for the run time cache to be `NULL` with the other fix then, so this change doesn't seem needed anymore.

SirSnyder: You guys are right, this part is superfluous. I will update the PR, thx.

tpunt:
Closing this since it was superseded by the changes made in #755 (which has now been merged).

tpunt:
I've reopened this since my attempt to refactor the static preparation code has failed. I'm still unsure whether introducing a new global for this is the best way, though. Do you see any other ways? Otherwise, can you update the PR to reflect the comments made on it, and I'll look into merging it. Thanks.

src/pthreads.h
```diff
@@ -122,6 +122,7 @@ ZEND_BEGIN_MODULE_GLOBALS(pthreads)
 	zval  this;
 	HashTable resolve;
 	HashTable *resources;
+        zend_bool prepare_static_members;
```
tpunt: nit: this should be tab indented

SirSnyder:
@tpunt unfortunately I see no other solution. I've updated the PR.

tpunt:
Merged in 8b10643, thanks.

Issue #706, opened 2017-05-28
Socket::read() normal read


userqq:
Hi. 
unlike to <a href="http://php.net/manual/en/function.socket-read.php">socket_read</a>, Socket::read() requires flags from <a href="http://php.net/manual/en/function.socket-recv.php">socket_recv</a> function (MSG_OOB, MSG_PEEK, MSG_WAITALL, MSG_DONTWAIT) and only default read mode (PHP_BINARY_READ) is possible.

any plans to add PHP_NORMAL_READ?

krakjoe:
There are plans to enhance Socket, but I don't have the time right now ...

SirSnyder:
Merged #828

PR #744, opened 2017-08-28
#706 - Socket::read() adds PHP_NORMAL_READ and PHP_BINARY_READ

SirSnyder:


classes/socket.h
```diff
@@ -237,12 +237,13 @@ PHP_METHOD(Socket, select) {
 PHP_METHOD(Socket, read) {
 	zend_long length = 0;
 	zend_long flags = 0;
+        zend_long type = PHP_BINARY_READ;
```
tpunt: nit: this should use tabs for indentation.

php_pthreads.c
```diff
@@ -324,7 +324,7 @@ PHP_MINIT_FUNCTION(pthreads)
 	INIT_CLASS_ENTRY(ce, "Socket", pthreads_socket_methods);
 	pthreads_socket_entry = zend_register_internal_class_ex(&ce, pthreads_threaded_entry);
 	pthreads_socket_entry->create_object = pthreads_socket_ctor;
-
+
```
tpunt: nit: unnecessary whitespace change.

SirSnyder:
@tpunt, new editor with working indentation ;-) files updated

src/socket.h
```diff
@@ -30,6 +30,9 @@ typedef struct _pthreads_socket_t {
 	zend_bool blocking;
 } pthreads_socket_t;
 
+#define PHP_NORMAL_READ 0x0001
+#define PHP_BINARY_READ 0x0002
```
dktapps: you might want to define Socket class constants for these. These won't be available as userland constants if ext/sockets isn't available.

SirSnyder:
Follow up on https://github.com/krakjoe/pthreads/pull/828

PR #745, opened 2017-08-28
Fix getting dependencies in AppVeyor builds

dktapps:


SirSnyder:
@dktapps I have no win build setup at hand right now. But I guess you've tested your changes and it works, right?

dktapps:
@SirSnyder it was tested as of August, yes. The AppVeyor build for the PR passes. Presumably it still works, I use the same configuration for CI builds on my own extensions.

SirSnyder:
Thanks!

PR #746, opened 2017-08-28
Bug/72 class constant visibility

SirSnyder:
This PR based on PR 741. The zend_class_entry constants_table contains zend_class_constant instead of zend_constant objects. The present behavior leads to segfaults. It is necessary to copy the ce and doc_comment, also the delayed build/preparation. If the changed function name is not ok, I will rename/revert the function name.

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -48,7 +48,7 @@ static  zend_trait_method_reference * pthreads_preparation_copy_trait_method_ref
 static void pthreads_prepared_resource_dtor(zval *zv); /* }}} */
 
 /* {{{ */
-static void pthreads_prepared_entry_static_members(zend_class_entry *candidate, zend_class_entry *prepared) {
+static void pthreads_prepared_entry_static_members(pthreads_object_t* thread, zend_class_entry *candidate, zend_class_entry *prepared) {
 	if (candidate->default_static_members_count) {
 		int i;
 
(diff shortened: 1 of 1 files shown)


src/copy.h
```diff
@@ -288,7 +288,7 @@ static inline zend_function* pthreads_copy_internal_function(zend_function *func
 static zend_function* pthreads_copy_function(zend_function *function) {
 	zend_function *copy = zend_hash_index_find_ptr(&PTHREADS_ZG(resolve), (zend_ulong)function);
 	
-	if (copy) {
+	if (copy && copy->op_array.run_time_cache == NULL) {
```
tpunt: This change should not be here.

tpunt:
@SirSnyder Please rebase this onto master, and then I'll merge it.

SirSnyder:
@tpunt I've updated the PR, please have a look. Thx!

tpunt:
@SirSnyder I've taken another look at this, but I don't seem to be able to reproduce the problem you're seeing using the master branch of pthreads and the PHP-7.2 branch of php-src. Assuming you can still see this problem, could you add a test with a reproducing script, so that I can better see what this fixes exactly? Also, another nitpick, but the whole body of the main change is indented with spaces (I only noticed this when viewing it in my editor) - can this be changed to use tabs please?

dktapps:
@tpunt this somehow appears to have landed on master in https://github.com/krakjoe/pthreads/commit/b879981e5742be10561246a0898f454a8176ea97

tpunt:
Thanks for letting me know. It looks like I pulled this change a while ago, but forgot that it was committed to my branch... I've reverted it for now, so that @SirSnyder can make the aforementioned changes first.

tpunt:
Merged with a  few changes and a test. Thanks!

SirSnyder:
Oh sorry, somehow I've missed this PR. Thanks @tpunt !

Issue #742, opened 2017-08-25
Retrieving finished Threaded task data when using thread Pool


Mista-Tea:
# Environment
Windows Server 2012
PHP 7.1.6 (cli)
PThreads: v3.1.6 (php_pthreads-3.1.6-7.0-ts-vc14-x64)

---

I apologize if this isn't the proper place to ask questions, but I'm on a severe time-crunch. Between outdated v2 examples floating around this repository, StackOverflow, and php.net mixed in with still-undocumented v3 changes, I'm having a hard time understanding what I need to do to retrieve data from a finished threaded task when using a pool.

I read your (tpunt's) article at https://www.sitepoint.com/parallel-programming-pthreads-php-fundamentals/ concerning the new Volatile class and some of the changes, but I am still ultimately confused.

My current setup involves using a Pool and an extended Worker class that can grab a connection to a MySQL database. I usually spin up thousands or more extended Threaded objects that access the worker's MySQL connection, grab some data that corresponds to the threaded job (like a user ID), run some algorithms, compute some scores, and then finish up.

What I'd like to be able to do is grab the scores from all of these jobs as they finish so that I can compute the final averages.

As the simplest example, take below where some work calculates a random number and assigns it to its $result field. 

```php
<?php
    class ExampleWork extends Threaded {
        public $result;
		
        public function run() {
            $this->result = rand();
        }
    }

    $pool = new Pool( 3, "Worker" );
	
    for ( $i = 0; $i < 12; $i++ ) {
        $pool->submit( new ExampleWork() );
    }
	
    while ($pool->collect());
    $pool->shutdown();
?>
```

How would I go about changing this example into a way for me to access each ExampleWork's $result as they finish/after all jobs finish?

Thanks for your time.

tpunt:
Hi!

A couple of points:
> PHP 7.1.6 (cli)

It is not safe to run pthreads on versions < PHP 7.2, since ZTS mode in PHP had flaws (creating race conditions). There is a chance you will experience some random segfaults.

> Between outdated v2 examples floating around this repository, StackOverflow, and php.net mixed in with still-undocumented v3 changes

To my knowledge, the PHP docs and the examples here should be updated to v3 - if there are some bits that aren't, then please do let me know. I can't do much about StackOverflow...

[...]

PR #747, opened 2017-08-29
#742 - simple (almost stupid) bug fix

SirSnyder:
Very simple bug fix

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -305,7 +305,9 @@ int pthreads_store_write(zval *object, zval *key, zval *write) {
 
 	if (Z_TYPE_P(write) == IS_ARRAY) {
 		if (!pthreads_check_opline_ex(EG(current_execute_data), -1, ZEND_CAST, IS_ARRAY) &&
-			!pthreads_check_opline_ex(EG(current_execute_data), -2, ZEND_CAST, IS_ARRAY)) {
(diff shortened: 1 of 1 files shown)


tpunt:
I'm not keen on this fix. I also looked into something similar initially, but didn't resolve it this way because it isn't robust enough.

Firstly, it still is just a hack that will not apply to all cases:
```php
$this->a = [];
$this->a[] = (array)[1];
$this->a->b = [];
$this->a->b[] = (array)[2];
$this->a->b->c = [];
$this->a->b->c[] = (array)[3];
$this->a->b->c->d = [];
$this->a->b->c->d[] = (array)[3]; // still a Volatile object
```

Secondly, when you begin looking back through the opline that far, you can no longer guarantee that you're still working off of the current statement. For example, with your fix, the following will be incorrect:
```php
$a = (array)[];
$this[] = []; // now an array
var_dump($this);
```

The second line will now be an array, rather than the expected `Volatile` object. This is because it has looked back 3 opcodes from the write operation, and has seen that one of them is an array cast, even though it is for a different expression.

I don't know what the solution is for this problem yet (or if there even is one), but it is not safe to look back more than two opcodes for an array cast.

tpunt:
Given the problems I mentioned above, I am closing this PR since it is not a suitable fix.

PR #770, opened 2017-10-23
Create composer.json

RAJESHINDIN:
Without composer.json file, people find it difficult to install the package in framworks like laravel. So, it will make the installation easy using composer


RAJESHINDIN:
Created composer.json file in root folder

rask:
This is an extension written in C, not a package that can be installed inside other PHP projects/packages.

bwoebi:
Exactly, no point in a composer.json here. Closing.

PR #774, opened 2017-11-02
fix compile error on 7.2

dktapps:
wrong version check, `GC_ADDREF` only exists in 7.3-dev. If I read correctly, https://github.com/krakjoe/pthreads/commit/edf81d3cc0f34c1fcb3f8f14435a9a122f6e482f was actually intended to fix 7.3, not 7.2.


tpunt:
Merged, thanks.

krakjoe:
Thanks for cleaning up there ... I'm unsure what I thought I was doing :D

PR #778, opened 2017-12-05
Ability to nest volatile objects

SirSnyder:
@tpunt @krakjoe Volatile objects are not suitable for nested and/or complex structures, because of the missing refcount. I hope the contained test clarifies the use case of this PR.

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -143,7 +143,7 @@ static inline zend_bool pthreads_store_is_immutable(zval *object, zval *key) {
 /* {{{ */
 int pthreads_store_delete(zval *object, zval *key) {
 	int result = FAILURE;
-	zval member;
+	zval member, *property = NULL;
 	pthreads_object_t *threaded = PTHREADS_FETCH_FROM(Z_OBJ_P(object));
 	zend_bool coerced = pthreads_store_coerce(threaded->store.props, key, &member);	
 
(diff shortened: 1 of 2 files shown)


SirSnyder:
@tpunt @krakjoe could someone of you please review this PR. Do you see any problems concerning performance? This PR is important for Volatile objects and I would like to merge it soon.

tpunt:
Ok, I've given it a review now. I don't see anything technically wrong with it, but the negative performance impact it brings is not good.

Using the following code:
```php
<?php

class A extends Thread
{
	public function __construct(Threaded $t)
	{
		$this->t = $t;
	}

	public function run()
	{
		for ($i = 0; $i < 10000; ++$i) {
			$this->t[] = function (){};
		}
	}
}

$t = new Threaded();
$a = new A($t);

for ($i = 0; $i < 10000; ++$i) {
	$t[] = function (){};
}

$a->start() && $a->join();
```

Times:
 - *Before* the patch: **0.056s**
 - *After* the patch: **1.242s**

Instructions:
 - *Before* the patch: **198,469,277**
 - *After* the patch: **10,398,134,419**

That's quite a degradation for normal `Threaded` objects, and weakens the purpose of immutability in pthreads (something [Joe wrote](http://blog.krakjoe.ninja/2015/09/addendum-letter-from-future.html) quite fondly about).

In its current state, I don't think it would be a good idea to merge the patch. But we can use this patch as a base, and optimise it to provide a solution that minimises the performance degradation.

I can take another look at this over the weekend.

SirSnyder:
I'm afraid I made a mistake. Hold on, I will push a fix tomorrow.

SirSnyder:
I've restricted `pthreads_store_sync()` in line https://github.com/krakjoe/pthreads/pull/778/files#diff-299af04dc20fc53f6c029d9f260211b6R377 to Volatiles only. @tpunt please repeat your benchmark.

tpunt:
I did wonder whether you meant to restrict the syncing to just `Volatile` objects only. This looks fine to me now, and gives a nice performance win on the following script (which was problematic in issue #786):
```php
<?php

class ThreadedClass extends Thread
{
    private $work;
    private $matrixSize;

    public function __construct($matrixSize) {
        $this->matrixSize = $matrixSize;
        $this->work = new Volatile();
        // $this->work = new Threaded(); // for better performance
    }

    public function pushWork($line) {
        $this->work[] = $line;
    }

    public function run() {
        for ($i = 0; $i < count($this->work); $i++) {
            for ($j = 0; $j < $this->matrixSize; $j++) {
                $this->work[$i][$j] = rand(1, 100);
            }
        }
    }
}

$matrix = $threads = [];
$matrixSize = 500;
$numThreads = 2;

for ($i = 0; $i < $matrixSize; $i++) $matrix[$i] = [];
for ($i = 0; $i < $numThreads; $i++) $threads[] = new ThreadedClass($matrixSize);
for ($i = 0; $i < $matrixSize; $i++) $threads[$i % $numThreads]->pushWork($matrix[$i]);
for ($i = 0; $i < $numThreads; $i++) $threads[$i]->start();
for ($i = 0; $i < $numThreads; $i++) $threads[$i]->join();
```

Feel free to merge.

SirSnyder:
With pleasure! How did you measure the number of instructions?

tpunt:
> How did you measure the number of instructions?

By using [Cachegrind](http://valgrind.org/docs/manual/cg-manual.html):
```
valgrind --tool=cachegrind -- php -dextension=modules/pthreads.so script.php
```

Issue #776, opened 2017-11-14
socket_recvfrom()/socket_sendto() equivalents?


dktapps:
Environment
==========
  * PHP: 7.2.0RC5
  * pthreads: 3.1.7dev (https://github.com/krakjoe/pthreads/commit/1c9787e5d918a3602661290ff5efe32ba3df74c8)
  * OS: Windows 10 (x64)

Summary
======
There don't appear to be any equivalents for the `socket_recvfrom()` and `socket_sendto()` functions in PHP sockets. This presents a problem when wanting to work with connectionless protocols (like UDP).

SirSnyder:
@dktapps could you test the PR, please?

PR #780, opened 2017-12-05
WIP: #776 - socket_recvfrom / socket_sendto support

SirSnyder:
Support of socket_recvfrom() and socket_sendto(), equal to php socket API


src/socket.c
```diff
@@ -696,4 +696,179 @@ void pthreads_socket_free(pthreads_socket_t *socket, zend_bool closing) {
...
+#endif
+	socklen_t			slen;
+	int					retval;
+	char				*address;
+	zend_string *recv_buf;
+
+	recv_buf = zend_string_alloc(len + 1, 0);
+
+	switch (threaded->store.sock->type) {
+		case AF_UNIX:
```
dktapps: this will need a check for win32

src/socket.c
```diff
@@ -696,4 +696,179 @@ void pthreads_socket_free(pthreads_socket_t *socket, zend_bool closing) {
...
+	struct sockaddr_un	s_un;
+	struct sockaddr_in	sin;
+#if HAVE_IPV6
+	struct sockaddr_in6	sin6;
+	char				addr6[INET6_ADDRSTRLEN];
+#endif
+	int	retval;
+
+	switch (threaded->store.sock->type) {
+		case AF_UNIX:
```
dktapps: same here

SirSnyder:
check for win32 added, @dktapps

dktapps:
I haven't had chance to test this yet, will take a look over the weekend.

dktapps:
Some issues I picked up on:

- The extension can't be compiled without PHP sockets. Use of `PHP_SOCKET_ERROR()` makes pthreads dependent on the `sockets` extension, as does usage of the below functions. Shouldn't that be `PTHREADS_SOCKET_ERROR()`?
```
socket.obj : error LNK2019: unresolved external symbol PHP_SOCKET_ERROR referenced in function pthreads_socket_recvfrom
socket.obj : error LNK2019: unresolved external symbol php_set_inet_addr referenced in function pthreads_socket_sendto
socket.obj : error LNK2019: unresolved external symbol php_set_inet6_addr referenced in function pthreads_socket_sendto
```
- MSVC complains about a bunch of unreferenced variable declarations that should probably be moved under their relevant cases.
```
ext\pthreads\src\socket.c(827): warning C4101: 'addr6': unreferenced local variable
```
- `sockaddr_un` is not defined on Windows, again the declarations could probably be moved under the AF_UNIX cases.

dktapps:
@SirSnyder I've created a couple of commits you might be interested in cherry-picking (commits 1a1f562cd38f326d5a00f5e314b8d7ba2ab084e9 and 60401fe65d79c36828dccad0cc5c8d99908c0255).

I've tested this and it seems to be working fine with my additional patches, but one problem I did spot:
```
socket_recvfrom($socket, $buf, 65535, 0, $addr, $port);
```
works, and I don't have to define `$buf`, `$addr` and `$port` separately (since they are set by `socket_recvfrom()` by reference), but
```
$socket->recvfrom($buf, 65535, 0, $addr, $port);
```
raises TypeErrors (`Fatal error: Uncaught TypeError: Argument 1 passed to Socket::recvfrom() must be of the type string, null given`), requiring those variables to be separately declared. Maybe consider allowing the reference parameters to be nullable.

dktapps:
Also, I haven't tested IPv6 or Unix socket yet.

src/socket.c
```diff
@@ -696,4 +696,172 @@ void pthreads_socket_free(pthreads_socket_t *socket, zend_bool closing) {
...
+		case AF_INET6: {
+			struct sockaddr_in6	sin6;
+			char				addr6[INET6_ADDRSTRLEN];
+
+			slen = sizeof(sin6);
+			memset(&sin6, 0, slen);
+			sin6.sin6_family = AF_INET6;
+
+			if (port == NULL) {
+				efree(recv_buf);
```
dktapps: shouldn't this be `zend_string_free()`? (it is also the same in PHP sockets though)

dktapps:
@SirSnyder please also take a look at https://github.com/dktapps/pthreads/commit/27ee665cb6d0d96f326f2b9cd46fa483d8b5ba53

src/socket.c
SirSnyder: Yes, you're right. This should be ```zend_string_free()``, changed.

SirSnyder:
Thanks for your commits @dktapps! Would you report the efree/zend_string_free issue to php/php-src? Otherwise I would do that.

dktapps:
@SirSnyder reported, thanks.

tests/socket-sentto-recvfrom-ipv6-udp.phpt
```diff
@@ -0,0 +1,64 @@
+--TEST--
+Test if socket_recvfrom() receives data sent by socket_sendto() via IPv6 UDP
+--CREDITS--
+Copied from php/php-src and adjusted, originally created by 
+Falko Menge <mail at falko-menge dot de>
+PHP Testfest Berlin 2009-05-09
+--SKIPIF--
+<?php
+if (substr(PHP_OS, 0, 3) == 'WIN') {
+	die('skip Not valid for Windows');
```
dktapps: This is valid if compiled with `--enable-ipv6`. The AppVeyor configure options might need some adjustments.

SirSnyder: Can you adjust AppVeyor to support IPV6?

dktapps: Sure, I'll create a PR. Also, some of these tests aren't working correctly (see https://travis-ci.org/krakjoe/pthreads/jobs/334677874#L800)

SirSnyder: Looks like a travis issue https://github.com/travis-ci/travis-ci/issues/8711

SirSnyder: For the time being fixed with eaa0a589f10138fae2cd47c4dd227f30b655c095

PR #781, opened 2017-12-05
WIP: socket class stub

SirSnyder:


SirSnyder:
Yes, by hand :) a generated one would be great!

PR #794, opened 2018-01-09
Remove remnants of PTHREADS_ALLOW_GLOBALS

dktapps:
This constant does not exist in pthreads v3 as of 3.1.7.

### DIFF
--- a/examples/stub.php
+++ b/examples/stub.php
@@ -52,11 +52,6 @@ define('PTHREADS_INHERIT_COMMENTS', 0x100000);
 define('PTHREADS_ALLOW_HEADERS', 0x1000000);
 
 /**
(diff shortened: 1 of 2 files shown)


tpunt:
This constant could also be removed from the source code (see [thread.h#L87](https://github.com/krakjoe/pthreads/blob/d814b0cf701a2a5fa93bb69f61d1c29178086e82/src/thread.h#L87)).

dktapps:
@tpunt ah, I didn't know that was still there. I only saw that it wasn't exposed to PHPland anymore by way of `php --re pthreads`. I'll remove the define if you want.

SirSnyder:
Thanks @dktapps !

PR #805, opened 2018-01-26
Fixed arginfo on some Socket methods

dktapps:


tpunt:
Looks good to me.

There also seems to be a few other incorrect things in this file.

The required arg count is incorrect for a few other methods:
 - `Socket_setOption` arginfo should be `3`
 - `Socket_select` arginfo should be `3`
 - `Socket_read` arginfo should be `1`

The commented prototypes for the following are also incorrect:
 - `Socket::select`
 - `Socket::send`

Would you like to fix these in this PR too? If not, I'll merge this and fix them later.

dktapps:
@tpunt sure, I'll update the PR.

dktapps:
I'm not sure about `Socket::select()`, `socket_select()` requires the first timeout parameter. But the parameter parsing allows it to not be specified, and the commented doc indicates that all of them are required. Very inconsistent.

tpunt:
@dktapps I'd be inclined to go with what the current implementation does (i.e. what the parameter parsing API specifies). That will likely be the most recently updated piece of code (the arginfo usually ends up going out of sync).

dktapps:
@tpunt I've updated the PR

tpunt:
Merged in 8028ec103b9293b7b86dedfdb75de1a400eaf8e0, thanks!

PR #807, opened 2018-01-29
enable IPv6 support when testing on AppVeyor and add IPv6 Windows tes…

dktapps:
…t (based from php-src)


SirSnyder:
Thanks @dktapps !

PR #808, opened 2018-01-29
Fail AppVeyor build if any tests fail

dktapps:
Prior to this, AppVeyor builds would still be marked as "succeeded" even if tests failed.


SirSnyder:
Thanks @dktapps !

dktapps:
You're welcome :)

PR #811, opened 2018-01-29
tiny fix for win32 IPv4 test

dktapps:
Windows adds newlines at the end of most socket error messages.

see https://ci.appveyor.com/project/krakjoe/pthreads/build/master.274/job/lwjo6yv8mi5vb79v#L743


tpunt:
@SirSnyder It may be best to keep future socket development in different branches, until they stabilise a bit more. Otherwise, if the socket changes need to be reverted (for whatever reason), a dozen commits will need to be reverted (messy business), instead of just one. It's also not good for master stability, which we need whilst there's no official pthreads release for PHP 7.2 yet :)

SirSnyder:
@tpunt you're right. Thought it could be easy going to add some socket tests. But well, it wasn't ^^. I would love to see a release for PHP 7.2. In two weeks the last release was two years ago after all.

Issue #816, opened 2018-02-01
Segfault when unserializing objects with recursive references in arrays


dktapps:
*If you know how to write phpt tests, please open a PR with environment information, and your test, this will cause CI to test your issue before a human is able to look*

Environment
==========

  * PHP: 7.2.1
  * pthreads: latest master and mixed others, also tested with #778 
  * OS: windows x64, ubuntu 17.10 x64

Summary
======
This bug is a little difficult to describe. I'll let the code explain itself.

Importantly here https://gist.github.com/dktapps/049208bb4f8651d7e3bf9c46d4727313#file-pthreads-serialize-segfault-php-L24 MUST contain more than 1 object with recursive reference or the crash does not occur.

If explicit serialize()/unserialize() is done, the bug does not occur. It appears to be something pthreads itself is doing differently regarding serialization.

Reproducing Code
==============
https://gist.github.com/dktapps/049208bb4f8651d7e3bf9c46d4727313

Expected Output
=============
```
scheduling... 0
scheduling... 1
scheduling... 2
scheduling... 3
scheduling... 4
scheduling... 5
scheduling... 6
scheduling... 7
scheduling... 8
scheduling... 9
scheduling... 10
scheduling... 11
scheduling... 12
scheduling... 13
scheduling... 14
scheduling... 15
scheduling... 16
scheduling... 17
scheduling... 18
scheduling... 19
ok
```

Actual Output
============
```
scheduling... 0
scheduling... 1
scheduling... 2
scheduling... 3
scheduling... 4
Segmentation fault (core dumped)
```

Backtrace (Windows, VS)
================
```
>	[Inline Frame] php7ts.dll!zend_mm_alloc_small(_zend_mm_heap *) Line 1273	C	Symbols loaded.
 	[Inline Frame] php7ts.dll!zend_mm_alloc_heap(_zend_mm_heap * heap, unsigned __int64) Line 1344	C	Symbols loaded.
 	php7ts.dll!_emalloc(unsigned __int64 size) Line 2433	C	Symbols loaded.
 	[Inline Frame] php7ts.dll!zend_string_alloc(unsigned __int64 len, int) Line 134	C	Symbols loaded.
 	php7ts.dll!zend_string_tolower(_zend_string * str) Line 2599	C	Symbols loaded.
 	php7ts.dll!zend_lookup_class_ex(_zend_string * name, const _zval_struct * key, int use_autoload) Line 912	C	Symbols loaded.
[...]

PR #817, opened 2018-02-02
Fixed missing null terminator on serialized string

dktapps:
https://github.com/krakjoe/pthreads/issues/816#issuecomment-362667495

This is probably harmless since a length is stored for this anyway, but still unwanted behaviour.


dktapps:
This problem appears in a couple of other places as well, will push fixes.

tpunt:
Given that an extra byte was specifically allocated, I'm going to assume it was meant for a NUL termination, so I've merged. Thanks.

PR #818, opened 2018-02-02
#816 - keep refcount as set by unserialize

SirSnyder:


SirSnyder:
A test for this case is still missing

SirSnyder:
I have added a test.

tpunt:
Looks good to me. I'm not quite sure why the refcount was reset in the first place. The setting of the `pzval` to null is a little redundant, since it will be set to null in the calling function (it will also be set to null by the calling function of the calling function... But perhaps that can be cleaned up later).

SirSnyder:
@dktapps nice to see
@tpunt thanks for the hint, I've removed the nulling

PR #819, opened 2018-02-02
Socket error handling

SirSnyder:
The return type of the following methods had to be removed from zend arg info, because these methods can return bool(`false`) in case of an error: 

- connect()
- read()
- write()
- send()
- select()

The two methods `recvfrom()` and `sendto()` can also return bool(`false`) in case of an error, but there were no explicit return types defined.

Each Socket object holds his own error state now. To query and clear the error state, the two new methods `getLastError()` and `clearError()` were added. Method `getLastError()` differs from the PHP equivalent `socket_last_error()` by one param. Param `clear` was added to query and clear error state in one single method call. The param is optional and `false` by default.

I've tried to mimic the PHP socket error behavior(without warnings) with a view to blocking/non-blocking mode as far as possible. `Socket::select()` should also not throw exceptions in case of an error. But this method is static and there is no global Socket error state. Therefore a optional param `error` was added to `Socket::select()` to retrieve a possible error.

Two *-win32 socket tests where removed. @dktapps could you test the both socket-sentto-recvfrom-* tests on windows, please?


classes/socket.h
```diff
@@ -66,22 +69,22 @@ ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_listen, 0, 1, _IS_BOOL, 0)
 	ZEND_ARG_TYPE_INFO(0, backlog, IS_LONG, 0)
 ZEND_END_ARG_INFO()
 
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_connect, 0, 2, _IS_BOOL, 0)
+ZEND_BEGIN_ARG_INFO_EX(Socket_connect, 0, 0, 2)
```
dktapps: this doesn't look like it needs changes?

classes/socket.h
```diff
@@ -111,18 +114,23 @@ ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_getHost, 0, 0, IS_ARRAY, 0)
 	ZEND_ARG_TYPE_INFO(0, port, _IS_BOOL, 0)
 ZEND_END_ARG_INFO()
 
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_select, 0, 3, IS_LONG, 0)
+ZEND_BEGIN_ARG_INFO_EX(Socket_select, 0, 0, 4)
```
dktapps: In the code the `sec` parameter still appears to be optional, so the required argcount should be 3.

classes/socket.h
```diff
@@ -215,7 +225,7 @@ PHP_METHOD(Socket, listen) {
 	pthreads_socket_listen(getThis(), backlog, return_value);
 } /* }}} */
 
-/* {{{ proto Socket Socket::accept([string class = self::class]) */
+/* {{{ proto Socket/bool Socket::accept([string class = self::class]) */
```
dktapps: the separator ought to be `|`.

src/socket.c
```diff
@@ -877,9 +930,35 @@ void pthreads_socket_sendto(zval *object, int argc, zend_string *buf, zend_long
...
+void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_value) {
+	pthreads_object_t *threaded =
+		PTHREADS_FETCH_FROM(Z_OBJ_P(object));
+
+	PTHREADS_SOCKET_CHECK(threaded->store.sock);
+
+	if(threaded->store.sock->error == SUCCESS) {
+		RETURN_FALSE;
+	}
+	ZVAL_LONG(return_value, threaded->store.sock->error);
```
dktapps: the `RETVAL_LONG` macro could be used here

src/socket.c
```diff
@@ -648,12 +687,15 @@ static int pthreads_sockets_from_fd_set(zval *sockets, fd_set *fds) /* {{{ */
...
 /* }}} */
 
-void pthreads_socket_select(zval *read, zval *write, zval *except, uint32_t sec, uint32_t usec, zval *return_value) {
+void pthreads_socket_select(zval *read, zval *write, zval *except, uint32_t sec, uint32_t usec, zval *errorno, zval *return_value) {
 	fd_set rfds, wfds, efds;
 	php_socket_t mfd = 0;
 	int result = SUCCESS, sets = 0;
 	struct timeval  tv;
 
+	zval_dtor(errorno);
```
tpunt: This should be [`zval_ptr_dtor`](https://phpinternals.net/docs/zval_ptr_dtor), otherwise it will leak on cyclic references.

src/socket.c
```diff
@@ -54,22 +61,29 @@
...
-	int eno = php_socket_errno(); \
-	char *estr = eno != SUCCESS ? \
-		php_socket_strerror(eno, NULL, 0) : \
-		NULL; \
-	zend_throw_exception_ex(spl_ce_RuntimeException, eno, \
-		"Error (%d): %s", eno, estr ? estr : "unknown"); \
-	if (eno != SUCCESS) { \
-		if (estr) { \
-			efree(estr); \
+#define PTHREADS_HANDLE_SOCKET_ERROR(eno) do { \
```
tpunt: Conventionally, parentheses should be placed around `eno` in the macro body.

src/socket.c
dktapps: Additionally this will segfault if the `errorno` parameter is not given.

SirSnyder: I've copied this part from https://github.com/krakjoe/pthreads/blob/d32079fb4a88e6e008104d36dbbf0c2dd7deb403/src/socket.c#L775. Is `zval_dtor` in `pthreads_socket_recvfrom` also faulty?

src/socket.c
SirSnyder: Parentheses added, should be fixed

SirSnyder:
Some remarks where fixed, thx @dktapps and @tpunt. Please have a look at it again.

src/socket.c
```diff
@@ -947,7 +947,7 @@ void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_
 	if(threaded->store.sock->error == SUCCESS) {
 		RETURN_FALSE;
 	}
-	ZVAL_LONG(return_value, threaded->store.sock->error);
+	RETURN_LONG(threaded->store.sock->error);
```
dktapps: this will not work as expected, did you mean `RETVAL_LONG`?

dktapps:
@SirSnyder from a non-technical perspective I don't see anything immediately wrong with it, except for `Socket::select()` will crash if the `error` reference parameter isn't specified.

src/socket.c
SirSnyder: Yep, you're right. `RETURN_LONG` contains a `return;`. My fault.

src/socket.c
SirSnyder: Ok. `zval_dtor` was replaced by `zval_ptr_dtor`. Maybe you could have a look at the original implementation of `socket_recvfrom` https://github.com/php/php-src/blob/master/ext/sockets/sockets.c#L1771 which uses `zval_dtor` too.

SirSnyder:
@dktapps `Socket::select()` won't crash anymore without `error` specified. I've added a null pointer check. Do you guys see any more issues? Otherwise I will merge this PR later today.

src/socket.c
tpunt: Ok, I'll have a look into this, thanks.

src/socket.c
```diff
@@ -54,20 +61,27 @@
...
+			php_socket_strerror((eno), NULL, 0) : \
+			NULL; \
+		zend_throw_exception_ex(spl_ce_RuntimeException, (eno), \
+			"Error (%d): %s", (eno), estr ? estr : "unknown"); \
+		if ((eno) != SUCCESS) { \
+			if (estr) { \
+				efree(estr); \
+			} \
 		} \
+		return; \
```
tpunt: I could be wrong, but won't this return statement prevent the `return_value` being set to `false` later on?
https://github.com/krakjoe/pthreads/blob/f15c95e65d52d9f3770094fea199b4ecdb036cae/src/socket.c#L741-L743

tpunt: Also in the `PTHREADS_SOCKET_ERROR` call sites.

tpunt:
I've had another look through the code, and cannot see anything immediately wrong with it (other than the comment I left). Assuming @dktapps doesn't notice any further problems, then feel free to merge. Also, more tests around the sockets functionality (testing both success and error paths) would be good.

src/socket.c
```diff
@@ -54,20 +61,27 @@
...
-	char *estr = eno != SUCCESS ? \
-		php_socket_strerror(eno, NULL, 0) : \
-		NULL; \
-	zend_throw_exception_ex(spl_ce_RuntimeException, eno, \
-		"Error (%d): %s", eno, estr ? estr : "unknown"); \
-	if (eno != SUCCESS) { \
-		if (estr) { \
-			efree(estr); \
+#define PTHREADS_HANDLE_SOCKET_ERROR(eno) do { \
+	if ((eno) != EAGAIN && (eno) != EWOULDBLOCK && (eno) != EINPROGRESS && (eno) != SOCK_EINVAL) { \
```
tpunt: `EWOULDBLOCK` was written to be conditionally used lower down. Will it always be available? If not, then it should be conditionally used here too.

src/socket.c
SirSnyder: In case of an exception, there is no need to set a return value or not?

SirSnyder:
@tpunt more tests will come

src/socket.c
SirSnyder: On modern unix systems with GNU C library `EWOULDBLOCK` is always available with same value of EAGAIN. On modern windows systems `EWOULDBLOCK` is available but Microsoft do not want to guarantee `EWOULDBLOCK`. I've changed that. If error no `EWOULDBLOCK` doesn't exist on win32, it will be defined.

src/socket.c
tpunt: Quite true. I overlooked that part...

PR #820, opened 2018-02-03
Fixed IPv6 tests not running when ext/sockets is not compiled

dktapps:


tpunt:
It looks like this change was bundled into commit b508e06e19671bf4bceae93f34071243ea5d95df, so I'm closing this. Thanks.

SirSnyder:
Yep, thanks @dktapps

PR #823, opened 2018-02-04
fixed #812 - missing reference on copied functions

### LINKED ISSUE #812: Crash when calling collect on Worker running closures.
Environment
==========

PHP 7.2.1 (cli) (built: Jan  4 2018 04:29:12) ( ZTS MSVC15 (Visual C++ 2017) x86 )
pthreads = 3.1.7dev
Windows 10 Version 10.0.16299 Build 16299

Summary
======

Run this code for a while.  For me it crashes in about 7 seconds.  If you comment out the ``($this->closure)();`` it'll run forever.

Reproducing Code
==============

```
class TestClosure extends Threaded implements Collectable {
  protected $closure;

  function __construct( $closure ) {
     $this->closure = $closure;
  }

  public function run() {
    ($this->closure)();
  }

};

$worker = new Worker();
$worker->start();
while (true) {
  $function = new TestClosure(function() {
    return 1;
  });
  $worker->stack($function);
  while($worker->collect());
  sleep(1);
}


```

Expected Output
=============

This code should not crash.  It should run forever.

Actual Output
============

Exited with code=3221226505 in 7.084 seconds

dktapps:
since these are inserted into the resolve HT for later reuse, the refcount should reflect that. Without this fix the refcount only accounts for the reference returned, and not the one stored in the HT.

This fixes #812.

### DIFF
--- a/src/copy.h
+++ b/src/copy.h
@@ -392,6 +392,7 @@ static zend_function* pthreads_copy_function(zend_function *function) {
 		copy = pthreads_copy_internal_function(function);
 	}
 
+	function_add_ref(copy);
 	return zend_hash_index_update_ptr(&PTHREADS_ZG(resolve), (zend_ulong) function, copy);
 } /* }}} */
 #endif


dktapps:
Some tests are failing in debug mode (leaks).

PR #824, opened 2018-02-09
Auto-generated Socket class stub

dktapps:
(as mentioned in #781)

This is generated from latest master, built on Ubuntu. Constant values might be different on Windows and/or might be missing in some cases.

This does not have any documentation.


SirSnyder:
Thanks @dktapps

PR #826, opened 2018-02-09
Socket: fix Unix socket test failure when ext/sockets is not present

dktapps:


SirSnyder:
Could you add leading backslashes(`\Socket`)?

dktapps:
@SirSnyder sure, if you like. This code is in the global namespace though so it doesn't make any difference.

SirSnyder:
@dktapps yes I know it's nitpicking but now it's consistent to other socket tests ;) Thanks!

PR #827, opened 2018-02-10
Don't cache closures set as members in resolve HT

SirSnyder:


tests/closures-as-members-repeated-reallocation.phpt
```diff
@@ -0,0 +1,31 @@
...
+    public function run() {
+        ($this->closure)();
+    }
+}
+
+$count = 0;
+$worker = new Worker();
+$worker->start();
+
+while ($count++ < 5) {
```
dktapps: I just ran this test case on an arm64 binary (without this bug fix) and it doesn't reproduce with a low number due to ???. I suggest increasing this to something like 20 (which does reproduce the crash).

dktapps:
This doesn't appear to be working - with a higher loop count as mentioned above, the test case fails with segfault.

The way I would hypothetically do this is to check the `ZEND_ACC_CLOSURE` flag on the function in `pthreads_copy_function()` and avoid using the HT if it is set. This solution does make it clearer what is going on elsewhere though, although it could also be achieved with an extra parameter to `pthreads_copy_function()`.

SirSnyder:
That's an odd behavior. This fix works perfectly well on my i5 machine(php 7.2). I use a loop of 100000 for tests. Without a fix, it crashes at third loop. I wanna avoid a extra parameter to  don't blow up logic in `pthreads_copy_function()`. Sure, there are other ways to solve and I'm open for them :) but first we have to find the differences why this fix is working on my machine and not on yours.

dktapps:
never mind me, I'm not sure what I was doing this morning... I ran clean builds with 100k loops and everything worked fine (debug and release both). Possibly failed to checkout the branch because I downloaded the test case manually to run on master.

[This](https://transfer.sh/LEadh/dont-cache-closures.patch) is the fix I wrote this morning after thinking this wasn't working, but I guess it's not needed.

SirSnyder:
@dktapps I like your fix and have taken it over. @tpunt @krakjoe what do you think about this solution to prevent seg faults?

dktapps:
relevant issue: #812

SirSnyder:
ping @tpunt @krakjoe any concerns? I would like to merge this PR within the next days.

tpunt:
What was the reason for the segfault in the first place? It is not clear to me why this would cause a segfault.

dktapps:
@tpunt this was discussed [here](https://github.com/krakjoe/pthreads/issues/812#issuecomment-362907898)

tpunt:
@dktapps Right, I vaguely recall that issue. I never got around to debugging exactly why the function was prematurely freed from the resolve HT, though. Did you or @SirSnyder find out the reason for this?

dktapps:
@tpunt the `zend_function` behind closures has its refcount decremented when the Closure encapsulating it goes out of scope. Since the functions in the resolve HT don't have an extra ref on the count for the HT itself, they get destroyed when the closure goes out of scope, causing a fault next time it's accessed.

As I mentioned in the issue, the bulletproof solution would be to add an extra ref to prevent things disappearing from the resolve HT, but this is complicated because the refs would have to be cleaned up, and previously it wasn't just funcs in the resolve HT.

tests/closures-as-members-repeated-reallocation.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+Testing closure members reallocation
+--DESCRIPTION--
+This test verifies that closures can be set as members repeatedly
+--FILE--
+<?php
+class TestClosure extends \Threaded implements \Collectable {
```
tpunt: nit: we don't have a Collectable interface anymore

tpunt:
Ok, I see now. This fix looks like a good approach to me then. Whilst incrementing the refcount upon the initial insertion would prevent the copying overhead, handling the garbage collection of closures would be difficult to do efficiently and without introducing memory leaks, so it's not really worth it.

SirSnyder:
Thanks for your support @dktapps and @tpunt for review

tests/closures-as-members-repeated-reallocation.phpt
SirSnyder: my fault, should have checked that

PR #829, opened 2018-02-10
Ensure proper win32 sockets 

SirSnyder:
I've copied windows_commons.h from php-src to be sure all basic consts and vars are defined on win32. @dktapps could you test this branch please? Does it compile and work properly? Do you see any issues?

PR #830, opened 2018-02-11
Fixed missing libraries for Windows sockets

dktapps:
I have no idea how this actually worked before now. @SirSnyder this fixes Windows build for #829  .

All tests pass when run locally on x64.


SirSnyder:
Thanks @dktapps !

SirSnyder:
At least the inet6 stuff was quiet not working in some cases. `pthreads_socket_string_to_if_index` returned always false.

dktapps:
Yeah, I didn't test it in detail, just checked why it wasn't compiling.

Issue #806, opened 2018-01-28
stdin/stdout/stderr constants not defined


dktapps:
Environment
==========

  * PHP: 7.2.1
  * pthreads: 3.1.7dev
  * OS: windows

Summary
======
This issue is pretty much the same complaint as seen in https://github.com/krakjoe/pthreads/issues/44 . I'm making this issue because this is a problem when wanting to use library code that needs these constants and subsequently crashes because they aren't defined.

These can be defined manually in userland to work around this problem, but that is extra boilerplate for every thread class created. Would it not be more sensible to have a solution for this in pthreads itself?

I see that @krakjoe mentions that not all threads created in userland will need these constants, but that is also true of lots of thread-unaware library code. For code that _does_ need it it's a headache when these are not present. I was unaware that these constants were not copied/recreated before looking at pthreads' source code.

Reproducing Code
==============
```php
<?php

$thread = new class extends \Thread{
    public function run(){
        var_dump(defined('STDOUT'));
    }
}

$thread->start(PTHREADS_INHERIT_NONE) && $thread->join();
```

Expected Output
=============
```
bool(true)
```

Actual Output
============
```
bool(false)
```

SirSnyder:
Referring to #44, since pthreads doesn't support any web server(http) sapi anymore, shouldn't it be ok to expose stdin/stdout/stderr to threads?

dktapps:
I would assume so. For example my application has a (blocking) threaded console reader which reads user input typed into the console.

Since these constants are not defined, I just workaround the issue by opening resource handles.  A simple userland workaround for this is to just:
```php
define('STDIN', fopen('php://stdin', 'rb'));
define('STDOUT', fopen('php://stdout', 'wb'));
define('STDERR', fopen('php://stderr', 'wb'));
```

[These](https://github.com/krakjoe/pthreads/blob/master/php_pthreads.c#L99) are the currently supported SAPIs.

dktapps:
According to http://php.net/manual/en/features.commandline.io-streams.php, only the CLI SAPI (and phpdbg, but that isn't mentioned) defines these constants, so I don't see a problem anyway. They can simply be recreated on the child thread if they were found to exist on the parent thread, no?

SirSnyder:
Merged as 8f4205d

Issue #583, opened 2016-03-30
Improvement: option to prevent inheritance of stdin/out/err


Xerkus:
Currently there is no way to prevent stdin/stdout/stderr from being inherited by threads

In my case application starts plugin host process and attaches to its stdin/stdout, it is not much of a problem unless child thread of my plugin host will try to output anything, then it will corrupt rpc channels. I would prefer to use socket based rpc server but application does not support that, not yet at least.

krakjoe:
I'm not sure if this is reasonable ... let me think ...

PR #831, opened 2018-02-12
Copy stdin, stdout and stderr at thread creation

SirSnyder:
This PR copies the streams stdin, stdout and stderr(if available) to new threads at creation and independently of start options. See #806 and #583


dktapps:
Since they are resources, I was under the impression they would be unsafe to copy from the origin thread? I thought it would be better to just recreate them on the child.

krakjoe:
Correct, this would be unsafe ...

There might be a way to do it by dup'ing the handles and using the streams API to reimport them in the new thread ... I'd go with that, if anything ...

CharlotteDunois:
@SirSnyder Is there any update on this?

SirSnyder:
Not yet, maybe in 3 or 4 weeks.

SirSnyder:
@krakjoe can you have a look at the last commit in this PR, is the implementation ok?

dktapps:
@SirSnyder I don't know any crashing condition, but this is inherently unsafe anyway since these are resources.

SirSnyder:
@dktapps from php vm perspective this should be ok. The resources are created per thread context and not shared between them. So there is no resource/memory dependency to another thread. What can happen are race conditions by reading/writing to the streams. Developers have to sync them by hand. I think I'll wait for that until `FileStream`s are implemented.

tpunt:
On the one hand, this seems fine to me. Multiple threads writing to the same output will obviously incur race conditions without any kind of synchronisation, so anyone using this should synchronise the access for a predictable output (and I don't see any ways for this to crash from PHP's VM from such race conditions). But on the other hand, there's probably a nicer way to expose writing in an atomic way (e.g. to write strings in full, implicitly serialising them one after another without having output interleaved). I'm not against this PR, though.

dktapps:
fwiw, the reason I originally wanted this was because I had some library code that did checks on `STDOUT` to check for ANSI colour code support (it disables colours on pipes). I didn't really put any consideration into how they would be used for actual stdio.

SirSnyder:
@tpunt I have had the idea to replace the STD* constants by FileStream objects from https://github.com/krakjoe/pthreads/tree/streams. But that was just a crazy idea. I would say this stuff can get merged.

SirSnyder:
Thanks to all involved!

dktapps:
This has caused some problems for PM on Windows. Threads hang on start until the user presses the enter key on the console. I'm still trying to gauge the scope of the issue and reproducing test case, but this is not bug free.

dktapps:
@sirsnyder it would be appreciated if you could revert the merge, it makes things difficult for PRs since I have to omit this commit to be able to test anything.

sirsnyder:
@dktapps reverted

PR #833, opened 2018-02-16
Rework of Socket::connect

SirSnyder:
- Parameter `port` can be optional on AF_UNIX sockets
- Added some tests

### DIFF
--- a/classes/socket.h
+++ b/classes/socket.h
@@ -69,7 +69,7 @@ ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_listen, 0, 1, _IS_BOOL, 0)
 	ZEND_ARG_TYPE_INFO(0, backlog, IS_LONG, 0)
 ZEND_END_ARG_INFO()
 
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_connect, 0, 2, _IS_BOOL, 0)
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_connect, 0, 1, _IS_BOOL, 0)
 	ZEND_ARG_TYPE_INFO(0, host, IS_STRING, 0)
 	ZEND_ARG_TYPE_INFO(0, port, IS_LONG, 0)
 ZEND_END_ARG_INFO()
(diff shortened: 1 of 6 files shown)


SirSnyder:
@dktapps could you do a review?

src/pthreads.h
```diff
@@ -138,6 +138,8 @@ ZEND_END_MODULE_GLOBALS(pthreads)
 #define PTHREADS_PG(ls, v) PTHREADS_FETCH_CTX(ls, core_globals_id, php_core_globals*, v)
 #define PTHREADS_EG_ALL(ls) PTHREADS_FETCH_ALL(ls, executor_globals_id, zend_executor_globals*)
 
+#define PTHREADS_THROW_MSG(errno, msg) zend_throw_exception_ex(spl_ce_RuntimeException, (errno), (msg))
```
tpunt: Not much is really gained by making this a macro, particularly in one of the main pthreads headers (since it is only used in the socket.c file).

src/socket.c
```diff
@@ -415,8 +425,8 @@ void pthreads_socket_connect(zval *object, zend_string *address, zend_long port,
 			struct sockaddr_un s_un = {0};
 
 			if (ZSTR_LEN(address) >= sizeof(s_un.sun_path)) {
-				/* throw */
-				return;
+				PTHREADS_THROW_MSG(0, "Path too long");
+				RETURN_FALSE;
```
tpunt: This `RETURN_FALSE` (and the other two above) could simply be `return;`, since the exception thrown makes the return value inaccessible.

tpunt:
Other than the couple of comments I made, I don't see a problem with this. I think it makes sense to mimic the sockets API in php-src where possible.

classes/socket.h
```diff
@@ -236,16 +236,17 @@ PHP_METHOD(Socket, accept) {
...
+/* {{{ proto bool Socket::connect(string host[, int port]) */
 PHP_METHOD(Socket, connect) {
 	zend_string *host = NULL;
 	zend_long port = 0;
+	int argc = ZEND_NUM_ARGS();
 
-	if (zend_parse_parameters(ZEND_NUM_ARGS(), "Sl", &host, &port) != SUCCESS) {
-		return;
+	if (zend_parse_parameters(argc, "S|l", &host, &port) != SUCCESS) {
+		RETURN_FALSE;
```
dktapps: This seems redundant, isn't it usual to just `return;` here? never see anything else do this

SirSnyder: Socket::connect defines an explicit return type, so a bool value is here required too.

src/pthreads.h
SirSnyder: Thought about later extensibility, but the more I think about it, I'll change that.

src/socket.c
SirSnyder: Yes, you're right.

SirSnyder:
@tpunt sockets API in php-src is the template I'm following. The details are tricky to find, I will need some time to complete.

SirSnyder:
Thanks @tpunt and @dktapps

PR #835, opened 2018-02-19
Fixed unnecessary nullable method signature BC breaks in PHP 7.1+

dktapps:
In PHP 7.1+, these resulted in the signatures becoming `?typehint`, which raises warnings when upgrading old code which overrides these methods. As it turns out, these parameters are not supposed to accept null at all (see the implementation).
These nullable flags are not necessary; it is sufficient to declare that only X parameters are needed.

This will break code written for pthreads 3.1.7dev which gave null explicitly to these methods; however the fact that these methods accept null is not documented anywhere and I think it is a bug.

This is backwards compatible with code which overrides these methods since parameter type widening allows this.


SirSnyder:
Yes, it's undocumented and buggy behavior. I think these changes are ok, also with a BC break.

SirSnyder:
@tpunt what's your opinion about the BC break? Ok, not ok?

dktapps:
@SirSnyder The 3.1.6 release doesn't have these issues since it was only compatible with PHP 7.0 anyway. The bug only exists in versions of pthreads compatible with PHP 7.1 (3.1.7dev only), which since it doesn't have a release is technically not a BC break anyway.

dktapps:
I've pushed an additional fix to address the `Pool::__construct()` nulled typehints noted in #838 .

tpunt:
Merged, thanks.

Issue #836, opened 2018-02-19
Worker::collect() should accept any callable, but only accepts \Closure objects


dktapps:
Environment
==========

  * PHP: 7.2.2
  * pthreads: https://github.com/krakjoe/pthreads/commit/2bc44069c1f3c83c866464b60345a333eacee598
  * OS: windows x64

Reproducing Code
==============
```php
function collector(Collectable $collectable) : bool{
	return $collectable->isGarbage();
}

class Work extends \Threaded{
	public function run(){
		sleep(1);
	}
}
$worker = new \Worker();
$worker->start();
$worker->stack(new Work());
while($worker->collect('collector'));
$worker->shutdown();
```

Expected Output
=============
nothing

Actual Output
============
```
Fatal error: Uncaught TypeError: Argument 1 passed to Worker::collect() must be an instance of Closure, string given in pthreads-test.php:38
Stack trace:
#0 pthreads-test.php(38): Worker->collect('collector')
#1 {main}
  thrown in pthreads-test.php on line 38
```

PR #838, opened 2018-02-19
Fixed some issues with stub method signatures

dktapps:
This updates the stub to match pthreads v3 properly, as per the arginfo of the extension. Some issues were discovered in the process (see #835 and #836), so method signatures have been adjusted to match the documentation where there are mismatches.

The formatting in this file is very inconsistent, I went to great pains to avoid reformatting it to avoid polluting the diff. If it's desirable to have a reformat in this PR, let me know, but please don't squash-merge. :)

### DIFF
--- a/examples/stub.php
+++ b/examples/stub.php
@@ -3,7 +3,7 @@
  * pthreads extension stub file for code completion purposes
  *
  * @author Lisachenko Alexander <lisachenko.it@gmail.com>
- * @version 2.0.0
+ * @version 3.0.0
  */
 
 /**
@@ -74,11 +74,12 @@ class Threaded implements Traversable, Countable, ArrayAccess, Collectable
      * Fetches a chunk of the objects properties table of the given size
      *
      * @param int $size The number of items to fetch
+     * @param bool $preserve Preserve the keys of members
      *
      * @link http://www.php.net/manual/en/threaded.chunk.php
      * @return array An array of items from the objects member table
      */
-    public function chunk($size) {}
+    public function chunk($size, $preserve = false) {}
 
     /**
      * {@inheritdoc}
@@ -110,7 +111,7 @@ class Threaded implements Traversable, Countable, ArrayAccess, Collectable
      *
      * @return bool(true) The referenced object can be destroyed
      */
-    public function isGarbage() {}
+    public function isGarbage() : bool{}
 
     /**
      * Tell if the referenced object is executing
@@ -132,7 +133,7 @@ class Threaded implements Traversable, Countable, ArrayAccess, Collectable
      * Merges data into the current object
      *
      * @param mixed $from The data to merge
-     * @param bool $overwrite Overwrite existing keys flag, by default true
+     * @param bool $overwrite Overwrite existing keys flag
      *
      * @link http://www.php.net/manual/en/threaded.merge.php
      * @return bool A boolean indication of success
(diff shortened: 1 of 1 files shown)


SirSnyder:
I've checked your changes and they are fine. It's ok if you reformat stub.php in a second commit. :)

examples/stub.php
```diff
@@ -444,14 +441,7 @@ class Pool
...
      */
-    public function __construct($size, $class, array $ctor = array()) {}
-
-    /**
-     * Shuts down all Workers, and collect all Stackables, finally destroys the Pool
-     *
-     * @link http://www.php.net/manual/en/pool.__destruct.php
-     */
-    public function __destruct() {}
+    public function __construct(int $size, ?string $class = null, ?array $ctor = array()) {}
```
tpunt: This doesn't look quite right to me. The `$class` and `$ctor` parameters should not be nullable (attempting to pass in `null` to either of them will cause ZPP to fail). I think the arg info for the constructor should be updated to reflect this as well. Also, the default class is `Worker`, according to [the implementation](https://github.com/krakjoe/pthreads/blob/2bc44069c1f3c83c866464b60345a333eacee598/classes/pool.h#L80).

examples/stub.php
```diff
@@ -376,17 +375,15 @@ public function shutdown() {}
...
 
     /**
-     * Removes the referenced object ( or all objects if parameter is null ) from stack of the referenced Worker
-     *
-     * @param Threaded $work Threaded object previously stacked onto Worker
+     * Removes the first task (the oldest one) in the stack.
      *
      * @link http://www.php.net/manual/en/worker.unstack.php
-     * @return int The new length of the stack
+     * @return Collectable|null The item removed from the stack
```
tpunt: Perhaps it would be best to use `Threaded` here, rather than `Collectable`, since we use `Threaded` for `Worker::stack`, as well as `Pool::submit` and `Pool::submitTo`.

tpunt:
Other than the couple of comments I made on there, this patch looks fine to me.

PHP also has return typing that could be used in the stubs, too.

examples/stub.php
dktapps: @tpunt this was generated by a script, so if this is wrong then the arginfo is likely wrong as well. Probably this is the same issue as seen in #835 .

dktapps:
@tpunt as above, this was mostly done with a script, typehints have been added (or omitted) as per the arginfo already in place.

examples/stub.php
dktapps: On a semi-related note, the existence of `Collectable` as an interface is confusing as it isn't consistently used... it's also only implemented by `Threaded` and does not need to be implemented in userland (Volatile descends from Threaded anyway). To be honest it seems superfluous.

examples/stub.php
tpunt: @dktapps I see. Would you like to submit a PR to rectify this too? If not, I'll do so later today.

dktapps: I can push a fix with #835 if you want, since it's the same subject.

examples/stub.php
tpunt: Agreed. Perhaps that's something that can be removed in the next major version.

tpunt:
@dktapps Ok. Since it isn't going to be possible to add return typing to the arg infos, I guess it will have to be left to the doc comments (that, or updated manually, which is less desirable).

tpunt:
Merged, thanks.

PR #828, opened 2018-02-10
706 - Socket::read() adds Socket::PHP_NORMAL_READ and Socket::PHP_BINARY_READ

SirSnyder:
This PR adds `$type` to Socket::read() and the both constants `Socket::PHP_NORMAL_READ` as well as `Socket::PHP_BINARY_READ` to mimic php-src Sockets API. Four more tests were added.

PR #839, opened 2018-02-19
Rework of sockets to mimic php-src sockets API

SirSnyder:
- Macros `PTHREADS_SOCKET_ERROR` and `PTHREADS_HANDLE_SOCKET_ERROR` extended by `msg` and `eno` for detailed exceptions
- Extended error handling
- Bug fix in Socket::getOption()
- Param `sec` of Socket::select() is mandatory. Valid values are `0`, `> 0` and `null`.
- Param `port` of Socket::bind() is optional, default value is `0`.
- Param `backlog` of Socket::listen() ist optional, default value is `0`.
- Added serveral tests


SirSnyder:
@dktapps @tpunt please review this PR. Looks like Appveyor has problems with phpsdk_deps, don't know why.

After this PR has been merged, more additions and tests has to come. PR #828, extension of get/setOptions(), importStream()/exportSocket() and getErrorMessage() are in preparation. Later on the IPV6 stuff.

src/socket.c
```diff
@@ -451,25 +434,29 @@ void pthreads_socket_read(zval *object, zend_long length, zend_long flags, zval
 
 	PTHREADS_SOCKET_CHECK(threaded->store.sock);
 
+	/* overflow check */
+	if ((length + 1) < 2) {
+		RETURN_FALSE;
+	}
```
tpunt: `zend_long` is a signed type, and overflow on signed types is undefined behaviour according to the C standard. PHP has a `ZEND_LONG_MAX` macro that could be used instead.

src/socket.c
```diff
@@ -731,26 +719,43 @@ void pthreads_socket_select(zval *read, zval *write, zval *except, uint32_t sec,
 	}
 
 	PHP_SAFE_MAX_FD(mfd, 0);
-	
-	if (sec || usec) {
+
+	/* If seconds is not set to null, build the timeval, else we wait indefinitely */
+	if (sec != NULL) {
```
tpunt: From what I can tell, `sec` can never be `NULL`, since it is not an optional parameter (the same check in php-src looks superfluous).

src/socket.c
```diff
@@ -731,26 +719,43 @@ void pthreads_socket_select(zval *read, zval *write, zval *except, uint32_t sec,
...
 		} else {
-			tv.tv_sec = sec;
+			tv.tv_sec = Z_LVAL_P(sec);
 			tv.tv_usec = usec;
 		}
+
+		tv_p = &tv;
+
+		if (sec == &tmp) {
+			zval_dtor(&tmp);
```
tpunt: Longs are not reference counted, and so there's no need to destroy them (the same operation in php-src looks superfluous - probably outdated code from PHP 5, when every type was reference counted).

classes/socket.h
```diff
@@ -183,7 +183,7 @@ PHP_METHOD(Socket, setOption) {
 	zend_long value = 0;
 
 	if (zend_parse_parameters(ZEND_NUM_ARGS(), "lll", &level, &name, &value) != SUCCESS) {
-		return;
+		RETURN_FALSE;
```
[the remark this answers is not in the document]
SirSnyder: `socket_set_option` has no explicit return type defined by `ZEND_BEGIN_ARG_INFO_EX`, but Socket::setOption uses `ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX`. Without return `false`, this leads to the following output:

`$socket = new Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, \Socket::SOL_TCP);
  $socket->getOption();` 

> Warning: Socket::getOption() expects exactly 2 parameters, 0 given in ...
>
>Fatal error: Return value of Socket::getOption() must be of the type integer, null returned in Unknown on line 0

src/socket.c
```diff
@@ -63,27 +63,25 @@
...
 	\
 } while(0)
 
-#define PTHREADS_SOCKET_ERROR(socket) do { \
-	int eno = php_socket_errno(); \
-	(socket)->error = eno; \
-	PTHREADS_HANDLE_SOCKET_ERROR(eno); \
+#define PTHREADS_SOCKET_ERROR(socket, msg, eno) do { \
+	(socket)->error = (eno); \
+	PTHREADS_HANDLE_SOCKET_ERROR((eno), (msg)); \
```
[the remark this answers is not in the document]
SirSnyder: Ok, I'm not familar with arguments in macros. Will change that.

src/socket.c
```diff
@@ -125,12 +123,12 @@ void pthreads_socket_set_option(zval *object, zend_long level, zend_long name, z
 void pthreads_socket_get_option(zval *object, zend_long level, zend_long name, zval *return_value) {
 	pthreads_object_t *threaded =
 		PTHREADS_FETCH_FROM(Z_OBJ_P(object));
-	socklen_t unused;
+	socklen_t unused = sizeof(int);
```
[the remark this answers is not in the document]
SirSnyder: If `unused` is not initialized, `Socket::getOption()` returns always `0` for the first call. `int` was copied from php-src, will change it to `zend_long`.

src/socket.c
```diff
@@ -246,40 +256,29 @@ void pthreads_socket_bind(zval *object, zend_string *address, zend_long port, zv
...
 #endif
-
 		case AF_INET: {
 			struct sockaddr_in *sa = (struct sockaddr_in *) sock_type;
 
 			sa->sin_family = AF_INET;
 			sa->sin_port = htons((unsigned short) port);
 			
 			if (!pthreads_socket_set_inet_addr(threaded->store.sock, sa, address)) {
-				/* throw */
```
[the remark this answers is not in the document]
SirSnyder: The corresponding exception will be thrown in `pthreads_socket_set_inet_addr`. We just have to `return` at this point. Same for `pthreads_socket_set_inet6_addr`.

src/socket.c
```diff
@@ -290,24 +289,20 @@ void pthreads_socket_bind(zval *object, zend_string *address, zend_long port, zv
...
-			RETURN_TRUE;
+			retval = bind(threaded->store.sock->fd, (struct sockaddr *)sa, sizeof(struct sockaddr_in6));
 		} break;
 #endif
+	}
 
-		default:
-			return;
-			/* throw */
+	if (retval != 0) {
```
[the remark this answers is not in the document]
SirSnyder: Yes, that would be clearer

src/socket.c
SirSnyder: Maybe the term `overflow check` is not right here. I've copied this snippet from php-src and think it is used to ensure `length` is not negativ and at least `1`. Because `length <= 0`would not make sense.

src/socket.c
SirSnyder: `sec` needs to be `null`able. The following applies to `Socket::select()` and `socket_select()`

- `sec` < 0 - Invalid argument
- `sec` = 0 - returns immediately
- `sec` > 0 - waits definitely
- `sec` = null - waits indefinitely

classes/socket.h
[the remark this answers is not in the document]
tpunt: Hmm, I do see the problem. This is also why the return type information macros are not particularly pervasive in php-src. I'm not so concerned what solution is chosen here (current solution looks fine).

classes/socket.h
```diff
@@ -249,13 +249,12 @@ PHP_METHOD(Socket, connect) {
...
-/* {{{ proto int|bool Socket::select(array &read, array &write, array &except [, int sec = 0 [, int usec = 0 [, int &error]]]) */
+/* {{{ proto int|bool Socket::select(array &read, array &write, array &except, int sec [, int usec = 0 [, int &error]]) */
 PHP_METHOD(Socket, select) {
-	zval *read, *write, *except, *errorno = NULL;
-	zend_long sec = 0;
+	zval *read, *write, *except, *sec, *errorno = NULL;
 	zend_long usec = 0;
 
-	if (zend_parse_parameters(ZEND_NUM_ARGS(), "a/!a/!a/!|llz/", &read, &write, &except, &sec, &usec, &errorno) != SUCCESS) {
+	if (zend_parse_parameters(ZEND_NUM_ARGS(), "a/!a/!a/!z/!|lz/", &read, &write, &except, &sec, &usec, &errorno) != SUCCESS) {
```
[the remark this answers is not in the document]
tpunt: Since the parameter isn't being typed, then it is perfectly valid to pass in `null` to it. Separation is only needed for by-reference passing.

src/socket.c
tpunt: That sounds correct - perhaps the comment should be removed and the condition updated to `length < 1`.

src/socket.c
tpunt: Right, I see. I didn't realise `sec` would be `NULL` (C null) if a PHP `null` was passed in to it (I assumed it would have been `Z_TYPE_P(sec) == IS_NULL`). My mistake...

classes/socket.h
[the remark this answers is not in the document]
SirSnyder: I'm going to remove the separation. What would you recommend, keeping `sec` as zval or moving back to `zend_long`?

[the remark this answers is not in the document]
SirSnyder: I think I messed up something. `zend_long` is just an int32_t, so it won't be possible to handle null with it, right?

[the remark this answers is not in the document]
tpunt: That's correct. The exclamation mark will enable for it to be nullable, but you won't be able to differentiate between passing `null` or `0` to the parameter. So it must be kept as a zval.

src/socket.c
SirSnyder: Good to know. I've removed the `zval_dtor` call.

SirSnyder:
@tpunt thanks for your review so far, any further notes?

src/socket.c
```diff
@@ -431,16 +411,19 @@ void pthreads_socket_connect(zval *object, int argc, zend_string *address, zend_
...
-			}
-
-			RETURN_TRUE;
+			retval = connect(threaded->store.sock->fd, (struct sockaddr *) &s_un,
+					(socklen_t)(XtOffsetOf(struct sockaddr_un, sun_path) + ZSTR_LEN(address)));
 		} break;
 #endif
 	}
+
+	if (retval != 0) {
```
tpunt: nit: `retval != SUCCESS`

SirSnyder: I've missed this line, sorry!

examples/stub.php
```diff
@@ -662,13 +662,13 @@ public function getOption(int $level, int $name) : int{}
...
 
-    public function listen(int $backlog) : bool{}
+    public function listen(int $backlog = 0) : bool{}
 
     public function accept($class = self::class){}
 
     public function connect(string $host, int $port = 0) : bool{}
 
-    public static function select(array &$read, array &$write, array &$except, int $sec = 0, int $usec = 0, int &$error = null){}
+    public static function select(array &$read, array &$write, array &$except, int $sec, int $usec = 0, int &$error = null){}
```
dktapps: This should be `?int $sec`

SirSnyder: Good catch, do you see any other issues?

dktapps: Nothing off the hop, this is the only thing I noticed.

tests/socket-select.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+    $socket->close();
+--EXPECTF--
+
+Warning: Socket::select() expects at least 4 parameters, 0 given in %s on line %d
+int(0)
+int(0)
+int(0)
+bool(false)
+int(22)
```
dktapps: This test does not work as expected on Windows (I'm getting all zeros, no errors) - not sure if this is a Windows-specific bug or what, haven't investigated in-depth yet.

tests/socket-getoption.phpt
```diff
@@ -0,0 +1,18 @@
...
+<?php
+    $socket = new Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, \Socket::SOL_TCP);
+    $socket->getOption();
+
+    try {
+        $socket->getOption("hello", "world");
+    } catch(Throwable $throwable) {
+        var_dump($throwable->getMessage());
+    }
+    var_dump($socket->getOption(SOL_SOCKET, SO_REUSEADDR));
```
dktapps: use of global ext/socket constants causes this test to fail without ext/sockets

tests/socket-connect-error.phpt
```diff
@@ -15,17 +15,24 @@ display_errors=1
...
 Warning: Socket::__construct() expects exactly 3 parameters, 2 given in %s on line %d
-
-Fatal error: Uncaught TypeError: Argument 1 passed to Socket::__construct() must be of the type integer, array given in %ssocket-connect-error.php:%d
-Stack trace:
-#0 %ssocket-connect-error.php(%d): Socket->__construct(Array, 1, 1)
-#1 {main}
-  thrown in %ssocket-connect-error.php on line %d
\ No newline at end of file
+string(%d) "Argument 1 passed to Socket::__construct() must be of the type integer, array given"
+string(%d) "Unable to create socket (%d): Address family not supported by protocol"
```
dktapps: This error message is different on Windows: 
```
005+ string(96) "Unable to create socket (10047): An address incompatible with the requested protocol was used.
005- string(%d) "Unable to create socket (%d): Address family not supported by protocol"
006+ "
```
Note the trailing newline as well (as seen before with Windows sockets).

tests/socket-setoption.phpt
```diff
@@ -0,0 +1,20 @@
...
+    $socket = new Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, \Socket::SOL_TCP);
+    $socket->setOption();
+
+    try {
+        $socket->setOption("hello", "world", new stdClass());
+    } catch(Throwable $throwable) {
+        var_dump($throwable->getMessage());
+    }
+    var_dump($socket->setOption(SOL_SOCKET, SO_REUSEADDR, 1));
+    var_dump($socket->getOption(SOL_SOCKET, SO_REUSEADDR));
```
dktapps: same again here - use of global ext/sockets constants causes this to fail without ext/sockets

tests/socket-getsockname.phpt
```diff
@@ -0,0 +1,45 @@
...
+    var_dump($inetSocket->getSockName(), $inetSocket->getSockName(false));
+
+    $inetSocket->close();
+
+    $address = sprintf("/tmp/%s.sock", uniqid());
+
+    if (file_exists($address))
+        die('Temporary file socket already exists.');
+
+    $unixSocket = new \Socket(\Socket::AF_UNIX,\Socket::SOCK_STREAM, 0);
```
dktapps: This fails on Windows (no UNIX sockets).

SirSnyder:
I will start a second branch for fixes

tests/socket-getsockname.phpt
SirSnyder: Forgot to split the test, fixed.

tests/socket-select.phpt
SirSnyder: On win32, a negative timeout seems to be handled like a zero timeout, without any error.

SirSnyder:
Thanks @tpunt and @dktapps for your support!

PR #840, opened 2018-02-21
New static method Socket::strerror()

SirSnyder:
Added missing method Socket::strerror(int $error):?string equal to php-src sockets API

### DIFF
--- a/examples/stub.php
+++ b/examples/stub.php
@@ -691,4 +691,6 @@ class Socket extends \Threaded
     public function getLastError(bool $clear = false){}
 
     public function clearError(){}
+
+    public static function strerror(int $error) : ?string{}
 }
--- a/src/socket.c
+++ b/src/socket.c
@@ -969,6 +969,15 @@ void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_
 	}
 }
 
+void pthreads_socket_strerror(zend_long error, zval *return_value) {
+	char *errstr;
+	errstr = php_socket_strerror(error, NULL, 0);
+
+	RETVAL_STRING(errstr);
+
+	efree(errstr);
+}
+
 void pthreads_socket_clear_error(zval *object) {
 	pthreads_object_t *threaded =
 		PTHREADS_FETCH_FROM(Z_OBJ_P(object));
--- a/classes/socket.h
+++ b/classes/socket.h
@@ -40,6 +40,7 @@ PHP_METHOD(Socket, getSockName);
 
 PHP_METHOD(Socket, getLastError);
 PHP_METHOD(Socket, clearError);
+PHP_METHOD(Socket, strerror);
 
 PHP_METHOD(Socket, close);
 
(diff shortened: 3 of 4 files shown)


src/socket.c
```diff
@@ -969,6 +969,15 @@ void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_
...
 
+void pthreads_socket_strerror(zend_long error, zval *return_value) {
+	char *errstr;
+	errstr = php_socket_strerror(error, NULL, 0);
+
+	RETVAL_STRING(errstr);
+
+	efree(errstr);
+}
+
```
SirSnyder: @tpunt do you see any problem here with the alloc(`php_socket_strerror`) / efree() or is this part fine? `php_socket_strerror` returns an `char*` in every case, so it shouldn't be necessary to check for null, I think.

examples/stub.php
```diff
@@ -691,4 +691,6 @@ public function close(){}
     public function getLastError(bool $clear = false){}
 
     public function clearError(){}
+
+    public static function strerror(int $error):?string {}
```
dktapps: This formatting is not consistent with elsewhere in the stub (`strerror(int $error) : ?string{}`)

src/socket.c
tpunt: This seems fine to me. `errstr` will always be ZMM-allocated if `NULL` is passed in as the buffer, and `RETVAL_STRING` will internally perform a copy the returned `char *` buffer, so `errstr` does indeed need to be freed afterwards.

tpunt:
I've given it a quick review, and it seems that this implementation deviates from the `socket_strerror` implementation [in php-src](https://github.com/php/php-src/blob/bff179888f6d247c3584150f0c729634b95c745b/ext/sockets/sockets.c#L603), causing some different outputs. For example:
```php
<?php

var_dump(
    Socket::strerror(-10001)
);

var_dump(
    socket_strerror(-10001)
);
```

Outputs:
```
string(21) "Unknown error: -10001"
string(12) "Unknown host"
```

I'm not entirely sure why the php-src implementation [does this](https://github.com/php/php-src/blob/bff179888f6d247c3584150f0c729634b95c745b/ext/sockets/sockets.c#L607-L609), or why it's specifically not for Windows. That should probably be looked into before merging this.

SirSnyder:
php-src moves error codes from `getaddrinfo()` [in space beneath -10000](https://github.com/php/php-src/blob/master/ext/sockets/sockaddr_conv.c#L42). Appropriate error codes are defined in [php_gai_strerror()](https://github.com/php/php-src/blob/master/main/network.c#L105) but not used. I would like to add these separate error codes with the implementation of `socket_addrinfo_*`

tpunt:
Ah, that makes more sense, thanks. This looks good to me then.

examples/stub.php
SirSnyder: Fixed, do you see any other issue?

SirSnyder:
Thanks @tpunt and @dktapps

PR #841, opened 2018-02-22
Fixed signatures of Worker::collect() and Pool::collect()

### LINKED ISSUE #836: Worker::collect() should accept any callable, but only accepts \Closure objects

dktapps:
This fixes #836 (incorrect arginfo, again ^.^) - the implementations of these methods accept any callable, but the arginfo demanded Closure objects (break from v2). Since these are documented as accepting `callable`, this should be treated as a bug.

Also fixes Pool::collect()'s optional parameter.


examples/stub.php
```diff
@@ -452,7 +452,7 @@ public function __construct(int $size, string $class = Worker::class, array $cto
      *
      * @link http://www.php.net/manual/en/pool.collect.php
      */
-    public function collect(callable $collector) {}
+    public function collect(callable $collector = null) {}
```
tpunt: The default is technically the built-in `Worker::collector` method (same with the `Worker::collect` stub). I'm not sure how it can really be represented in the method signature, though. `[Worker::class, 'collector']` isn't accurate, since it's not a statically callable method, and I'm not sure `[$this, 'collector']` is much better either. Thoughts? Otherwise, perhaps it should just be mentioned in the doc comment instead.

tpunt:
Technical side of the patch looks good to me.

examples/stub.php
dktapps: I wasn't sure on this either. A doc comment is probably the best option.

tpunt:
Merged, thanks!

PR #844, opened 2018-02-27
Added static method Socket::createListen()

SirSnyder:
- Static method `createListen`
- One test


src/socket.c
```diff
@@ -1041,6 +1041,56 @@ void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_
...
+
+	created = PTHREADS_FETCH_FROM(Z_OBJ_P(return_value));
+	sock = created->store.sock;
+
+#ifndef PHP_WIN32
+	if ((hp = php_network_gethostbyname("0.0.0.0")) == NULL) {
+#else
+	if ((hp = php_network_gethostbyname("localhost")) == NULL) {
+#endif
+		zval_ptr_dtor(return_value);
```
SirSnyder: @tpunt is `zval_ptr_dtor` used correctly at this point and below?

SirSnyder:
@dktapps @tpunt please review

src/socket.c
tpunt: This looks fine to me. Technically, `zval_dtor` could be used, since we know that the newly created socket object will not contain a cycle (so there's no need to place it into the GC root buffer). But it doesn't really matter.

src/socket.c
```diff
@@ -1041,6 +1041,56 @@ void pthreads_socket_get_last_error(zval *object, zend_bool clear, zval *return_
 	}
 }
 
+void pthreads_socket_create_listen(zend_long port, zend_long backlog, zval *return_value) {
+	struct sockaddr_in  la;
+	struct hostent		*hp;
+	pthreads_object_t	*created;
+	pthreads_socket_t	*sock;
+
+	object_init_ex(return_value, pthreads_socket_entry);
```
tpunt: This can return `FAILURE`. It's not immediately clear to me if any such paths exists where this can fail (from a quick look through the source code), but it may be worthwhile checking its return value for failure just in case I've missed some psychotic edge case.

classes/socket.h
```diff
@@ -164,6 +166,7 @@ zend_function_entry pthreads_socket_methods[] = {
 	PHP_ME(Socket, close, Socket_void, ZEND_ACC_PUBLIC)
 	PHP_ME(Socket, getLastError, Socket_getLastError, ZEND_ACC_PUBLIC)
 	PHP_ME(Socket, clearError, Socket_void, ZEND_ACC_PUBLIC)
+	PHP_ME(Socket, createListen, Socket_void, ZEND_ACC_PUBLIC|ZEND_ACC_STATIC)
```
dktapps: This doesn't look right. A separate arginfo needs to be defined for this.

PR #845, opened 2018-02-28
Added method Socket::shutdown()

SirSnyder:
- Method `shutdown()` added
- one test added


SirSnyder:
@dktapps please review

tests/socket-shutdown.phpt
```diff
@@ -0,0 +1,32 @@
...
+<?php
+    if(!method_exists(\Socket::class, 'shutdown')) {
+        die('skip.. Socket::shutdown() doesn\'t exist');
+    }
+--FILE--
+<?php
+    $socket = new \Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, 0);
+    $socket->shutdown();
+
+    var_dump($socket->shutdown(4)); // invalid - false
```
dktapps: wouldn't an exception be more appropriate here?

classes/socket.h
```diff
@@ -42,6 +42,9 @@ PHP_METHOD(Socket, getLastError);
 PHP_METHOD(Socket, clearError);
 PHP_METHOD(Socket, strerror);
 
+#ifdef HAVE_SHUTDOWN
```
dktapps: where is this macro defined?

classes/socket.h
```diff
@@ -399,6 +411,19 @@ PHP_METHOD(Socket, strerror) {
 	pthreads_socket_strerror(error, return_value);
 } /* }}} */
 
+#ifdef HAVE_SHUTDOWN
+/* {{{ proto bool Socket::shutdown(int how) */
+PHP_METHOD(Socket, shutdown) {
+	zend_long how_shutdown = 0;
+
+	if (zend_parse_parameters(ZEND_NUM_ARGS(), "l", &how_shutdown) != SUCCESS) {
```
tpunt: Is there any reason in particular why we don't default this parameter to `2`? I would assume that most of the time when we shutdown a socket, we'd like to close it for both reads and writes, so following PHP's `socket_shutdown` function seems like it would make sense.

src/socket.c
```diff
@@ -589,6 +589,23 @@ void pthreads_socket_close(zval *object, zval *return_value) {
 	threaded->store.sock->fd = PTHREADS_INVALID_SOCKET;
 }
 
+#ifdef HAVE_SHUTDOWN
+void pthreads_socket_shutdown(zval *object, zend_bool how_shutdown, zval *return_value) {
```
dktapps: the signature here is incorrect - `how_shutdown` should be `zend_long` not `zend_bool`.

src/socket.h
```diff
@@ -46,6 +46,9 @@ void pthreads_socket_read(zval *object, zend_long length, zend_long flags, zend_
 void pthreads_socket_write(zval *object, zend_string *buf, zend_long length, zval *return_value);
 void pthreads_socket_send(zval *object, zend_string *buf, zend_long length, zend_long flags, zval *return_value);
 void pthreads_socket_close(zval *object, zval *return_value);
+#ifdef HAVE_SHUTDOWN
+void pthreads_socket_shutdown(zval *object, zend_bool how_shutdown, zval *return_value);
```
dktapps: same as above (signature is incorrect)

tests/socket-shutdown.phpt
```diff
@@ -0,0 +1,32 @@
+--TEST--
+Test of Socket::shutdown() - testing params and separate shutdown of read and write channel
+--DESCRIPTION--
+Test that creating and closing sockets works as expected on all platforms (gh issue #798)
```
dktapps: this description might need fixing

dktapps:
Experiencing test failures:
[socket-shutdown.log](https://github.com/krakjoe/pthreads/files/1774514/socket-shutdown.log)

tests/socket-shutdown.phpt
dktapps: On Windows, this causes an `ENOTCONN` error to be emitted, which causes a `RuntimeException` to be thrown. I assume that in Winsock2, socket state is checked prior to checking the argument value [citation needed].

SirSnyder: @dktapps I slightly modified the test from bind()/listen() to connect(). Can you check the test on a windows machine?

PR #846, opened 2018-03-05
Fixed leaking pthread_mutexattr_t on Windows, closes #822

### LINKED ISSUE #822: Windows memory issues
Environment
==========

  * PHP: 7.2.2
  * pthreads: https://github.com/krakjoe/pthreads/commit/866ed9b671bbae9acd9207e6d9d0a1fee890addf
  * OS: windows 10 x64

Summary
======
Run the below code for a while on Windows. Watch the memory usage in TaskManager.

This issue does not occur when I tested the same code on Ubuntu 17.10; it appears to be Windows-specific.

Given 10 minutes or so the memory usage reaches ~1GB on my machine.

`memory_get_usage()` on both threads does not give any indication that there is a problem (both report roughly static memory usages).

Reproducing Code
==============
```php
<?php

ini_set('memory_limit', '256M');
gc_enable();

class Test extends Threaded{}

$worker = new class extends Worker{
	public function run(){
		ini_set('memory_limit', '256M');
		gc_enable();
	}
};
$worker->start();

while(true){
    $function = new Test();

    $worker->stack($function);
    while($worker->collect());
}
```

Expected Output
=============
Memory usage should remain stable, as it does on Ubuntu.

Actual Output
============
Memory usage steadily grows the longer the script is running.

dktapps:
This bug occurs due to implementation differences in pthread-w32 vs other platforms. In glibc, [`pthread_mutexattr_destroy` is a no-op]( https://code.woboq.org/userspace/glibc/nptl/pthread_mutexattr_destroy.c.html#__pthread_mutexattr_destroy), so not calling it did no harm, but in pthread-w32 it [actually does something important](https://github.com/GerHobbelt/pthread-win32/blob/master/pthread_mutexattr_destroy.c#L47), i.e. freeing memory.

This bug was discovered using Visual Studio memory profiler. It may affect other platforms depending on the libc in use.

I haven't written tests for this because observing it from an automated standpoint to watch for leaks is tedious and takes a long time. However, this fix has been tested manually with the test script in #822 and is confirmed to fix the leak for Windows.

### DIFF
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -39,7 +39,9 @@ pthreads_monitor_t* pthreads_monitor_alloc() {
 #else
 	pthread_mutexattr_settype(&at, PTHREAD_MUTEX_RECURSIVE_NP);
 #endif
-	if (pthread_mutex_init(&m->mutex, &at) != 0) {
+	int ret = pthread_mutex_init(&m->mutex, &at);
+	pthread_mutexattr_destroy(&at);
+	if (ret != 0) {
 		free(m);
 		return NULL;
 	}


dktapps:
The implementation detail here is that `pthread_mutexattr_t` is merely a struct in regular POSIX implementations, but is a pointer to a `pthread_mutexattr_t_` in pthreadw32. The memory allocated for the structure is allocated on the heap by `pthread_mutexattr_init()` and as such needs to be freed. Thus the missing destroy() call causes the memory to leak.

SirSnyder:
Thanks @dktapps !

PR #860, opened 2018-04-24
Guard against returning uninitialized variables in switch-cases

dktapps:
VS2017 code analysis complains about "Using uninitialized memory 'retval'". While these cases should theoretically never get hit, they shouldn't be returning uninitialized variables anyway.


SirSnyder:
Thanks @dktapps !

PR #861, opened 2018-04-24
Add test for crash when modifying static array after thread start

dktapps:
I haven't been able to find the root cause of this issue yet, but it stems from pthreads' handling of copied static arrays.
When pthreads copies a static array, modifying the array on the parent thread triggers heap corruption and other nasties for reasons I am yet to fully understand.

The test case in the PR demonstrates the bug. It needs to be noted that:
- the crash does not occur unless the `PTHREADS_INHERIT_CLASSES` flag is enabled (hence, it is to do with the way static arrays are being handled)
- the crash occurs in GC when removing members of the array (`Test::$array = [];` works, also `unset(Test::$array[0]);`)
- the crash only appears to occur if `Test::$array` is modified while the thread is running (hence the use of a `Worker` to demonstrate)
- the crash does not occur if the static array is populated with scalar values or arrays, only objects.

### DIFF
--- a/tests/statics-modify-copied-object-array.phpt
+++ b/tests/statics-modify-copied-object-array.phpt
@@ -1,0 +1,45 @@
+--TEST--
+Test static object arrays work correctly when modified after thread start
+--FILE--
+<?php
+
+class Test{
+	public static $array = [];
+
+	public static function init(int $i){
+		while(--$i > 0){
+			self::$array[] = new stdClass();
+		}
+	}
+}
+
+Test::init(3);
+
+$thread = new Worker;
+$thread->start(PTHREADS_INHERIT_CLASSES);
+
+var_dump(Test::$array);
+Test::$array = [];
+var_dump(Test::$array);
+
+echo "gc start\n";
+gc_collect_cycles();
+echo "gc end\n";
+$thread->shutdown();
+echo "script end\n";
+
+?>
+--EXPECTF--
+array(2) {
+  [0]=>
+  object(stdClass)#1 (0) {
+  }
+  [1]=>
+  object(stdClass)#2 (0) {
+  }
+}
+array(0) {
+}
+gc start
+gc end
+script end


dktapps:
It appears this is to do with the array serialization happening on the wrong thread.

pthreads serializes and unserializes statics on the child thread. The problem with this is that serialization can cause objects to be modified (for example reallocating the properties table when it doesn't exist), which then cause memory errors later on when they try to be destroyed on the main thread.

The solution to this appears to be to do all static serialization on the main thread before thread start, and then unserialize on the child.

A similar issue is in effect with the exception-handler-options test failure - see https://github.com/pmmp/pthreads/commit/0c28adb3f6f2538c61caadc8a4c90fd36195d1e6 for a solution.

Obviously a similar solution for statics would be rather more complex, and I really don't want to go down this avenue, but I see no other solution to the problem.

SirSnyder:
I've spent some days with research, debugging and tests and come to same conclusion. My approach was, to initially clean copy the objects to be serialized but even that was error prune. As an example, `zend_array_dup` allocates a new Hashtable though, but the containing `zvals` are shared (also their `zend_refcounted` object) and their refcounter incremented. If you just replace the `zval` with a new one, it leaks at shutdown, cause of the wrong refcount. A call of `zval_ptr_dtor` adds the `zval` of Thread 1 to the GC (as root object) of Thread 2 and that is one reason (of several possible) for the shutdown segfault.

At the moment I assume that I will take over your approach.

dktapps:
My solution is a nasty one. I'd love to see a better one if you have any ideas.

I'm dicing with the thought of an intermediary pthreads_store (like the property store, but inaccessible to user code) which things would be fed through from the main thread. That's a nasty solution also, but less nasty than others I had in mind.

dktapps:
I also experimented with switching TSRM contexts, but didn't get anywhere with that (and barely have an idea what I'm doing there, so it's dangerous territory).

dktapps:
Read/write is thread local, so that isn't an issue. There just needs to be an intermediary to copying statics when classes are copied to the child thread (on thread start and on worker task submit).

SirSnyder:
Jep, these are thread local. But what bothers me, would be the gap between serialization in `pthreads_prepare_parent` and unserialization later on in the new thread. In this time, there can happen a lot. Static properties may change, classes were added and also small potential for race conditions. With a proper vm handler, pthreads could track static properties continously. I'm still thinking about that.

dktapps:
In all honesty, the selective inheritance of statics (inherit primitives and arrays) is confusing and a royal pain in the ass. Real thread-locals do not behave this way, so I do not see why they should in pthreads. Instead I think copying should be confined to only _default_ static members, and not assigned values. This would kill several birds with one stone, at the expense of a BC break.

dktapps:
Additionally this may also be a problem with local static variables, but I haven't tested this.

dktapps:
From what I could see, anything where the serialization happened in the wrong context could potentially be affected. Constants would also be, but there are no ways to define complex constants which would reproduce this bug. Local static variables also - they exhibit the same behaviour as class statics wrt. copying, so I would expect they would be impacted by this bug too, but I'd need to test that.

SirSnyder:
A first draft https://github.com/krakjoe/pthreads/commit/00779ab492552ff86e2784b02d11de5625c5ce24. Works very well with your attached statics-modify-copied-object-array test. Three tests are failing because of the BC break.

dktapps:
One thing I don't understand on further investigation. pthreads is _already_ only copying the `default_static_members_table`, which is apparently _not_ default static members (as far as I can tell). Bug in php-src or no?

PR #863, opened 2018-05-04
Don't silence SOCK_EINVAL errors

dktapps:
Silently returning false on invalid arguments is incredibly annoying and frustrating because it conceals bugs. This pull request changes the `PTHREADS_HANDLE_SOCKET_ERROR` macro to throw `RuntimeException` when receiving an `EINVAL` error.

This change makes several bugs in tests apparent, which will be fixed in later commits.

### DIFF
--- a/tests/socket-select-win32.phpt
+++ b/tests/socket-select-win32.phpt
@@ -8,6 +8,7 @@ if (substr(PHP_OS, 0, 3) != 'WIN') {
 --FILE--
 <?php
     $socket = new Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, \Socket::SOL_TCP);
+    $socket->bind("127.0.0.1", 19132);
     $socket->listen();
 
     try {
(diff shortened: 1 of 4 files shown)


tests/socket-select-win32.phpt
```diff
@@ -8,6 +8,7 @@ if (substr(PHP_OS, 0, 3) != 'WIN') {
 --FILE--
 <?php
     $socket = new Socket(\Socket::AF_INET, \Socket::SOCK_STREAM, \Socket::SOL_TCP);
+    $socket->bind("127.0.0.1", 19132);
```
dktapps: Not sure if 19132 is an acceptable port - I just automatically typed it in because this is the port I most commonly use in my project. If it should be changed, let me know.

dktapps:
It might also be worthwhile adding higher-level detection for invalid calls on sockets - as we saw here, calling `listen()` on an unbound socket will raise a cryptic `InvalidArgumentException` instead of something more sensible - `listen()` raises exceptions like this because you can't `listen()` on an unbound socket.

dktapps:
Annoyingly most cases where this actually raises exceptions are not due to invalid arguments, however `EINVAL` still shouldn't be silenced because it hides bugs.

PR #864, opened 2018-05-04
Improvements & fixes for Socket::shutdown() PR

dktapps:
@SirSnyder this one's for you - some suggested alterations of my own for #845

### DIFF
--- a/classes/socket.h
+++ b/classes/socket.h
@@ -141,7 +141,7 @@ ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_strerror, 0, 1, IS_STRING, 1)
 ZEND_END_ARG_INFO()
 
 #ifdef HAVE_SHUTDOWN
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(Socket_shutdown, 0, 1, _IS_BOOL, 0)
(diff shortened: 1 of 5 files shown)


classes/socket.h
```diff
@@ -414,9 +414,14 @@ PHP_METHOD(Socket, strerror) {
 #ifdef HAVE_SHUTDOWN
 /* {{{ proto bool Socket::shutdown(int how) */
```
tpunt: This prototype becomes outdated with this change.

classes/socket.h
```diff
@@ -414,9 +414,14 @@ PHP_METHOD(Socket, strerror) {
...
+	zend_long how_shutdown = SHUT_RDWR;
 
-	if (zend_parse_parameters(ZEND_NUM_ARGS(), "l", &how_shutdown) != SUCCESS) {
+	if (zend_parse_parameters(ZEND_NUM_ARGS(), "|l", &how_shutdown) != SUCCESS) {
+		RETURN_FALSE;
+	}
+
+	if (how_shutdown != SHUT_RD && how_shutdown != SHUT_WR && how_shutdown != SHUT_RDWR) {
+		zend_throw_exception(spl_ce_InvalidArgumentException, "Invalid shutdown type", 0);
 		RETURN_FALSE;
```
tpunt: Since an exception is being thrown, we can change `RETURN_FALSE` to simply be a `return;`, since the return value won't be usable after an exception has been thrown.

dktapps:
updated, thanks!

SirSnyder:
Many thanks @dktapps !

PR #865, opened 2018-05-04
Sockets: Improve consistency when dealing with invalid arguments

dktapps:
- Use `InvalidArgumentException` where appropriate instead of `RuntimeException` and/or returning false.
- Fixed undefined behaviour when a negative `usec` value was given to `Socket::select()`.
- Fixed compiler warning due to above point (casting signed to unsigned).
- Make `Socket::select()` behave consistently across all platforms with regards to timeouts.


tpunt (approved):
Looks good to me!

dktapps:
@tpunt with regards to the previously-mentioned problems with return values on not enough arguments, I think it makes sense to instead use `zend_parse_parameters_throw()` on the `Socket` API.
That way we get `ArgumentCountError` instead of this stupid problem with warnings and useless returns.

`throw`ing is also more consistent with userland code, for example:
```php
<?php

function test($a, $b, $c, $d){

}

\Socket::select();
test();
```
emits:
```
Warning: Socket::select() expects at least 4 parameters, 0 given in test.php on line 7

Fatal error: Uncaught ArgumentCountError: Too few arguments to function test(), 0 passed in test.php on line 8 and exactly 4 expected in test.php:3
Stack trace:
#0 test.php(8): test()
#1 {main}
  thrown in test.php on line 3
```

If that's something worth having, it could also go into this PR. I don't think it makes sense to stick with the legacy warnings.

dktapps:
to be honest I also don't really understand why `throw`ing is opt-in, given that it automatically applies to user code anyway. 🤷‍♂️

src/socket.c
```diff
@@ -388,7 +388,7 @@ void pthreads_socket_connect(zval *object, int argc, zend_string *address, zend_
 			struct sockaddr_in sin = {0};
 
 			if (argc != 2) {
-				zend_throw_exception_ex(spl_ce_RuntimeException, 0, "Socket of type AF_INET requires 2 arguments");
+				zend_throw_exception_ex(spl_ce_InvalidArgumentException, 0, "Socket of type AF_INET requires 2 arguments");
```
dktapps: This maybe ought to be `ArgumentCountError` instead of `InvalidArgumentException`, but `WRONG_PARAM_COUNT` only throws exceptions in strict mode - not sure how to deal with this

dktapps:
The changes in this PR conflict with #863 - if that's merged before this, this will need further alterations (to fix tests).

PR #874, opened 2018-06-06
Allow Volatile::extend() to work as expected

dktapps:
`Threaded` objects are often poorly suited for runtime class extension because of their immutability characteristics. Classes which are not `Threaded` need to be adapted to extend `Threaded` (for example, overwriting a coerced array on a runtime-extended class crashes).

Because `Volatile` descends from `Threaded`, `Volatile::extend()` exists, but does not work as might be expected - it behaves exactly the same as `Threaded::extend()`.

This PR changes `Threaded::extend()` to apply the scope it was called from, instead of always applying the `Threaded` parent.

A possible alternative to this could be to simply move `extend()` to `Volatile` completely, since as I said above `Threaded` isn't suited to code that doesn't know it's `Threaded` anymore due to immutability.

### DIFF
--- a/classes/threaded.h
+++ b/classes/threaded.h
@@ -320,7 +320,7 @@ PHP_METHOD(Threaded, extend) {
     if (is_final)
         ce->ce_flags = ce->ce_flags &~ ZEND_ACC_FINAL;
 
-	parent = zend_get_executed_scope();
+	parent = zend_get_called_scope(EG(current_execute_data));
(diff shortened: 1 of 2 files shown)


SirSnyder:
To keep `Volatile` and `Threaded` consistent, I prefer the way of this PR.

SirSnyder:
Thanks @dktapps !

PR #876, opened 2018-06-08
Fixed a crash with GC related to unstack()

dktapps:
pthreads worker stack maintains a linked-list of items to be executed. When an item is removed from the stack with unstack(), it is freed, but the item currently running on the stack does not know about this, and still has to this destroyed task in its `->next` member. This running task is then later added to the GC stack, where [this](https://github.com/krakjoe/pthreads/blob/master/src/stack.c#L84) happens. The GC-stacked item still has a ptr to the destroyed task in its `->next` member, so this loop then tries to free memory which has already been freed.

The cause of this was found by @MCMrARM, send him your love.

This fixes the described bug by NULL-ing the `->next` member of items added to the end of the stack in `pthreads_stack_add_item()`.

### DIFF
--- a/src/stack.c
+++ b/src/stack.c
@@ -102,6 +102,7 @@ static inline void pthreads_stack_add_item(pthreads_stack_t *stack, pthreads_sta
 		item->prev = stack->tail;
 		stack->tail = item;
 	}
+	item->next = NULL;
 	stack->size++;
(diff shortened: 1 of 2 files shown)


dktapps:
The CI failures are not related to the pull request changes.

PR #878, opened 2018-07-01
Fixed infinite iteration on Threaded objects in PHP 7.3

dktapps:
This bug was caused by changes made in https://github.com/php/php-src/commit/d7f2dc4ec651628e10213625db6aee3559e214a9 .
`zend_hash_move_forward_ex()` no longer sets the given `HashPosition` to `HT_INVALID_IDX` when it reaches the end of the HT, which causes [this](https://github.com/krakjoe/pthreads/blob/99a86fab88ede56a7ae0bd3e70249425caf5ad05/src/object.c#L5) to always be true.

There are probably better solutions for this, but this was the simplest and least invasive fix I could come up with. It is also backwards-compatible with PHP 7.2.

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -1030,6 +1030,9 @@ void pthreads_store_reset(zval *object, HashPosition *position) {
 
 	if (pthreads_monitor_lock(threaded->monitor)) {
 		zend_hash_internal_pointer_reset_ex(threaded->store.props, position);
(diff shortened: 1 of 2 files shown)


dktapps:
Seems this has issues when iterating on empty Threaded objects - they'll always yield one `NULL` before a foreach exits. Please don't merge this yet if it's seen :)

Issue #560, opened 2016-02-01
Numeric/string array index


RobDonovan:
PHP: 7.0.2
Pthreads: 3.1.5
OS: Ubuntu

Took me a while to find this problem in my code. From this test/example to show the problem, it looks bad that I am using a numeric and a string, but in my actual app when the array was created it was using a getnextseq() function that returned a numeric, but later on I was receiving the seq via a socket buffer to read/unset, and then it was a string.

Since in PHP numerics and strings are loosely typed, they are treated the same, so this took a bit of investigating, as I didnt spot it.

Example:

```
class sharedArray extends Threaded {
        public function run(){}
}

class socketThread extends Thread
{
        public function __construct($sharedArray)
        {
                $this->sharedArray = $sharedArray;
        }

        public function run()
        {

                $idx = 1;
                $this->sharedArray[$idx] = 'data';

                var_dump($this->sharedArray);

                $idx = '1';
                unset($this->sharedArray[$idx]);

                var_dump($this->sharedArray);

        }
}

$sharedArray = new sharedArray();

$socketThreads = new socketThread($sharedArray);
$socketThreads->start();
$socketThreads->join();
```

The 'unset' doesnt work, since I'm guessing because internally it looks different to pthreads.

However, this would work in standard PHP:

```
        $idx = 1;
        $array[$idx] = 'data1';

        $idx = '1';
        unset($array[$idx]);

        var_dump($array);
```

Not sure if you can do something about that internally or not, or if its something I/people need to be aware of.

For now, I can just cast the return result of my getNextSeq() to a string, so its uniform in my app.

JesusTheHun:
Well, implicit cast for object and arrays are a known issue/feature (depends on how you see it) in PHP. 
Pthreads extension definitely implement its own set/get handlers with pthreads object.
About if it intentionally doesn't cast index in order to copy PHP's behavior, that is to Joe to say.
However I agree with you, it must be fixed or known.

krakjoe:
This has pretty low urgency ... it's not being ignored ;)

SirSnyder:
Added in f55ffa9250f3856c5b542f074b45cd56c5d2db82

PR #882, opened 2018-07-15
Updates for PHP 7.3

dktapps:
This is mostly just fixing issues inserting persistent strings into non-persistent hashtables. In some cases I have replaced them with interned strings, others just needed copying appropriately.

There are also fixes for iteration.

There is a BC break in the behaviour of Volatile object keys which I believe is caused by https://github.com/php/php-src/pull/3351. I haven't taken the time to try and fix this yet.


tests/volatile-arrays.phpt
```diff
@@ -54,7 +56,7 @@ object(Threaded)#%d (%d) {
       }
     }
   }
-  [0]=>
+  ["0"]=>
```
dktapps: @SirSnyder are you sure we want to ignore this instead of correcting it? It is a BC break after all.

SirSnyder: @dktapps nope, my fault. I assumed this would be the new and intended behavior of php arrays. But it's not, will have a look at it.

SirSnyder:
@dktapps the last commit brings pthreads back to be consistent with php 73 arrays, at least I hope so ;-) Do you see any issues with the fix? This could also be a possible solution for #560

dktapps:
@SirSnyder Looks good to me 👍

SirSnyder:
@tpunt @krakjoe are the Hashtable changes/fixes ok?

dktapps:
This isn't synced with the latest changes, I haven't had time to update it  yet.

dktapps:
@CharlotteDunois I haven't touched this since the 7.3 alphas. I'll take another look at it soon when I find time.

SirSnyder:
I'm busy with a huge feature for pthreads in the last few weeks. Would be cool if you can take care of it @dktapps

src/prepare.c
```diff
@@ -595,7 +615,7 @@ static inline void pthreads_prepare_constants(pthreads_object_t* thread) {
 						case IS_DOUBLE: Z_DVAL(constant.value)=Z_DVAL(zconstant->value); break;
 						case IS_STRING: {
 #if PHP_VERSION_ID >= 70300
-							Z_STR(constant.value)=Z_STR(zconstant->value);
+							Z_STR(constant.value)= zend_string_new(Z_STR(zconstant->value));
 #else
 							ZVAL_NEW_STR(&constant.value, zend_string_new(Z_STR(zconstant->value)));
 #endif
```
dktapps: This block probably does not need the if/else.

src/store.c
```diff
@@ -1030,15 +1039,30 @@ void pthreads_store_reset(zval *object, HashPosition *position) {
 
 	if (pthreads_monitor_lock(threaded->monitor)) {
 		zend_hash_internal_pointer_reset_ex(threaded->store.props, position);
+		if (zend_hash_has_more_elements_ex(threaded->store.props, position) == FAILURE) { //empty
+			*position = HT_INVALID_IDX;
```
dktapps: This is duct tape and should probably be fixed better later.

SirSnyder: Yes, it's not a beauty but it is working for now. And who knows when they will change this in php again.

dktapps:
I've run some tests and this seems to be ready to r&r. I'm sure more bugs will roll in later on.

@SirSnyder I think this is good to merge.

SirSnyder:
@dktapps that's nice to hear. Two days back I've played with debug builds and two tests have mem leaks. The one I've investigated is in `pthreads_copy_literals` and was probably caused by https://github.com/php/php-src/commit/e70618aff6f447a298605d07648f2ce9e5a284f5#diff-8f41738bb0ceb7bf7c9882a4b399ab6e. I have no write permissions for this branch anymore, so I will push the fix direct to master.

src/prepare.c
SirSnyder: yes, you're right

PR #890, opened 2018-08-12
Fixed bug in pthreads_preparation_copy_trait_alias()

dktapps:


SirSnyder:
@dktapps , could you add a test please?

dktapps:
I'm not 100% sure how to test this bug - I spotted it while writing updates for PHP 7.3 and pulled a fix. The code is obviously wrong, but I'm not 100% sure how it would manifest itself in PHP code. I'll try and write a test later.

SirSnyder:
Ok, I will add a small test. Thanks for your PR!

PR #891, opened 2018-08-12
Allocate user class methods inherited from internal classes on arena,…

### LINKED ISSUE #881: Memory leak of zend_internal_function from inheritance
Environment
==========

  * PHP: 7.2
  * pthreads: master
  * OS: Windows 10 x64

Summary
======
When stopping threads, copied internal functions are not freed. These are not known by the zend MM due to being allocated by `calloc()` and do not get freed when the thread stops, unlike user functions. As a result, starting and stopping lots of threads will cause memory to be lost.

It appears that this bug only comes into effect when inheritance is used (see below).

Reproducing Code
==============
Leaking
```php
<?php

while(true){
	$t = new class extends \Worker{};
	$t->start(PTHREADS_INHERIT_NONE);
	$t->shutdown();
}
```
Works as expected:
```php
<?php

while(true){
	$t = new \Worker; //no inheritance
	$t->start(PTHREADS_INHERIT_NONE);
	$t->shutdown();
}
```

Expected Output
=============
memory usage should remain stable

Actual Output
============
memory usage goes up forever

dktapps:
… fixes #881 

This is in line with how inheritance usually works (https://github.com/php/php-src/blob/master/Zend/zend_inheritance.c#L66).

### DIFF
--- a/src/copy.h
+++ b/src/copy.h
@@ -372,7 +372,7 @@ static inline zend_function* pthreads_copy_user_function(zend_function *function
 
 /* {{{ */
 static inline zend_function* pthreads_copy_internal_function(zend_function *function) {
-	zend_function *copy = calloc(1, sizeof(zend_internal_function));
+	zend_function *copy = zend_arena_alloc(&CG(arena), sizeof(zend_internal_function));
 	memcpy(copy, function, sizeof(zend_internal_function));
 	return copy;
 } /* }}} */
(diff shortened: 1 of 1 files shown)


SirSnyder:
After intensive consideration, I can not find any side effects. Therefore I will merge this PR.

SirSnyder:
Thanks @dktapps

PR #911, opened 2018-12-30
Streams support

SirSnyder:
This is a thread-safe port of the php streams layer, including wrappers, transports and filters. Resources were replaced by objects and the following new classes introduced.

* `Streams`
* `File`
* `Stream`
* `StreamContext`
* `StreamWrapper`
* `StreamFilter`
* `StreamBucket`
* `StreamBucketBrigade`
* `pthreads_user_filter`
* `FileStream`
* `SocketStream`

See [stub.php](https://github.com/krakjoe/pthreads/blob/6a88bdcdc5b439568b3708e2b81be1c406046f16/examples/stub.php). A `File` class definition in global scope will be a huge BC break though. Maybe we will find another name or introduce a namespace, whatever that looks like.

Predefined wrappers
* php
  * temp
  * memory
  * output
  * stdin
  * stdout
  * stderr
  * fd/
  * filter/
* file
* glob
* data
* ftp
* http

Predefined filters
* `string.rot13`
* `string.toupper`
* `string.tolower`

Todo
* I've no idea why travis fails. Maybe you have an idea.
* Tests
* more Tests
* Windows build and testing

What's next? Removal of Todos for sure. I would like to publish this in a timely manner as prerelease in a separate branch to get audience and feedback. Possibly with other BC breaking changes like #906 and the behavior of static members initialization.

On top of this PR a port of the openssl "ssl" wrapper will follow.

@tpunt @dktapps Thoughts, questions?

**Big thanks to the team behind php-src!**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .project
+*.lo
--- a/classes/filestream.h
+++ b/classes/filestream.h
@@ -1,0 +1,918 @@
+/*
+  +----------------------------------------------------------------------+
+  | pthreads                                                             |
+  +----------------------------------------------------------------------+
+  | Copyright (c) Joe Watkins 2018                                       |
+  +----------------------------------------------------------------------+
+  | This source file is subject to version 3.01 of the PHP license,      |
+  | that is bundled with this package in the file LICENSE, and is        |
+  | available through the world-wide-web at the following url:           |
+  | http://www.php.net/license/3_01.txt                                  |
+  | If you did not receive a copy of the PHP license and are unable to   |
+  | obtain it through the world-wide-web, please send a note to          |
+  | license@php.net so we can mail you a copy immediately.               |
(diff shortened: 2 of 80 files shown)


tpunt:
Great work @SirSnyder! I'll try to get around to reviewing the code some time soon, as well as give some feedback on the API exposed.

SirSnyder:
I understand the concerns, it will be a lot more code to care about. Nevertheless it is a great opportunity for pthreads, we can build many things up on streams. I heard two years ago for the first time, that resources should be replaced. To be honest, I think that will never happen and end like php 6 in the distant future. That task is huge, it's massive. Beside the core, API, every bundled and external extension dealing with resources have to replace them too. The db extensions are no fun, really. Even if I'm wrong, which would be great, implementation and data handling must be thread-safe. Objects only are not enough.

PR #913, opened 2019-01-03
refresh file list

remicollet:


package.xml
```diff
@@ -267,8 +269,6 @@
 	<file name="php_pthreads.c" role="src" />
 	<file name="php_pthreads.h" role="src" />
 	<file name="README.md" role="doc" />
-	<file name=".travis.yml" role="doc" />
-	<file name="appveyor.xml" role="doc" />
   </dir>
```
remicollet: There is no value to distribute these files in the pecl archive.

SirSnyder:
Thanks @remicollet !

PR #915, opened 2019-01-15
Added support for igbinary serializer

### LINKED ISSUE #834: Feature request: Compatibility with igbinary extension
Environment
==========

  * PHP: 7.2.2
  * pthreads: https://github.com/krakjoe/pthreads/commit/d882b1c3b1c1f6aa025d4fcd3e876c98da3741cb
  * OS: Windows x64

Summary
======
As we know pthreads serializes any non-thread-safe object set as a member of a thread-safe one. In some applications (such as mine) this serialization poses problems for performance and also for memory usage.

This issue is a request to allow pthreads to use user-defined serializers, or in some other way enable optional internal usage of the [igbinary](https://github.com/igbinary/igbinary) extension. igbinary in many scenarios can be faster than the PHP serializer, and also uses significantly less memory for larger objects (because it does not try to be human readable).

This can be done around by explicitly serializing using igbinary in userland, but it would be nice if pthreads had a mechanism to optionally use this internally (by php.ini directives?). Since pthreads only needs to be able to represent an object as a string for copying between threads, it should play with igbinary very nicely.

I'm not sure if this is something that needs to be raised on the igbinary repository as well (I don't know how this would be implemented internally, only that it can be done - session is a precedent).

dktapps:
[igbinary](https://github.com/igbinary/igbinary) is faster and uses less memory, at the cost of readability. This is perfect for pthreads.

This code is not very pretty and could probably use some cosmetic enhancement. However, this works fine (apart from some behavioural quirks) so I'm pulling this to get feedback.

Pass `--with-pthreads-igbinary` at compile-time to enable it.

CI does not currently test with igbinary enabled - that's a TODO.

This closes #834 .

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -7,6 +7,9 @@ PHP_ARG_WITH(pthreads-sanitize, whether to enable AddressSanitizer for pthreads,
 PHP_ARG_WITH(pthreads-dmalloc, whether to enable dmalloc for pthreads,
 [  --with-pthreads-dmalloc   Enable dmalloc for pthreads], no, no)
 
+PHP_ARG_WITH(pthreads-igbinary, whether to enable igbinary support for pthreads,
+[  --with-pthreads-igbinary   Enable igbinary serializer for pthreads], no, no)
+
 if test "$PHP_PTHREADS" != "no"; then
 	AC_MSG_CHECKING([for ZTS])   
 	if test "$PHP_THREAD_SAFETY" != "no"; then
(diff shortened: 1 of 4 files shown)


SirSnyder:
@dktapps I've pushed a code change direct to your PR, but it's just a suggest. The feature to create a separate branch with a pull request didn't worked. Let me know what you think about the change. It's is untested, would be greate if you could test it ;-) @tpunt what do you think about igbinary support? I'm +1 in favor.

dktapps:
Looks good to me off the hop. I'll add igbinary-enabled test configurations soon so we can see it run on CI.

tpunt:
I don't have any complaints about this, but do you have any benchmarks to show noticeable performance differences? And I presume there are no semantic differences from the PHP side?

krakjoe:
Do that, don't introduce more serial data, it doesn't make much sense.

Cheers
Joe

krakjoe:
If no one done it by weekend after this one, I'll make a start.

Cheers
Joe

dktapps:
Sounds good to me. I'll close this for now.

Issue #919, opened 2019-01-29
Permit overriding join() to intercept thread shutdowns triggered by pthreads itself


dktapps:
Environment
==========

  * PHP: 7.2.13
  * pthreads: 5eb80c0c691aa81e0d235bdd37f6f30b633c433e
  * OS: w10 x64

Summary
======
I would like to be able to have my custom `join()` function called when the thread context tries to be destroyed. Currently this isn't possible because [pthreads_join](https://github.com/krakjoe/pthreads/blob/99a86fab88ede56a7ae0bd3e70249425caf5ad05/src/object.c#L335) is called directly, without giving the user chance to react to it.

Reproducing Code
==============
```php
$t = new class extends \Thread{

	private $shutdown = false;

	public function run(){
		$this->synchronized(function(){
			while(!$this->shutdown){
				$this->wait();
			}
		});
	}

	public function join(){
		$this->synchronized(function(){
			$this->shutdown = true;
			$this->notify();
		});
	}
};

$t->start();
unset($t);
die("OK");
```

Desired output
=============
```
OK
```

Actual Output
============
Nothing - the script just hangs at the `unset()` call as pthreads tries to fruitlessly join it.

PR #922, opened 2019-02-01
invoke user defined join()

SirSnyder:
@dktapps does that correspond to your issue #919?

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -332,6 +332,7 @@ void pthreads_base_free(zend_object *object) {
 			if ((PTHREADS_IS_THREAD(base)||PTHREADS_IS_WORKER(base)) &&
 				pthreads_monitor_check(base->monitor, PTHREADS_MONITOR_STARTED) &&
 				!pthreads_monitor_check(base->monitor, PTHREADS_MONITOR_JOINED)) {
+				pthreads_call_user_join(base);
 				pthreads_join(base);
 			}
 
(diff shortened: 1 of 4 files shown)


dktapps:
It appears to, yes. Haven't looked at this in close detail yet though. :thumbsup:

src/object.c
```diff
@@ -358,6 +359,48 @@ void pthreads_base_free(zend_object *object) {
...
+				call.fcc.object = &object->std;
+				call.fcc.calling_scope = object->std.ce;
+				call.fcc.called_scope = object->std.ce;
+				call.fcc.function_handler = join;
+
+				zend_call_function(&call.fci, &call.fcc);
+			}
+		}
+	} zend_catch {
+		pthreads_monitor_add(object->monitor, PTHREADS_MONITOR_ERROR);
```
dktapps: isn't this case used for when the thread itself errored, as opposed to the parent thread crashing trying to stop it?

src/object.c
```diff
@@ -358,6 +359,48 @@ void pthreads_base_free(zend_object *object) {
...
+	pthreads_call_t call = PTHREADS_CALL_EMPTY;
+	zval zresult;
+
+	if (pthreads_monitor_check(object->monitor, PTHREADS_MONITOR_ERROR)) {
+		return;
+	}
+
+	ZVAL_UNDEF(&zresult);
+
+	pthreads_monitor_add(object->monitor, PTHREADS_MONITOR_RUNNING);
```
dktapps: perhaps not what you intended?

src/object.c
```diff
@@ -358,6 +359,48 @@ void pthreads_base_free(zend_object *object) {
...
+		}
+	} zend_catch {
+		pthreads_monitor_add(object->monitor, PTHREADS_MONITOR_ERROR);
+	} zend_end_try();
+
+	if (Z_TYPE(zresult) != IS_UNDEF) {
+		zval_ptr_dtor(&zresult);
+	}
+
+	pthreads_monitor_remove(object->monitor, PTHREADS_MONITOR_RUNNING);
```
dktapps: Is this necessary? I was under the impression that this would already have happened by this point.

SirSnyder: sorry, had copied the whole stuff from another function to see if it's working. now it's clean.

PR #943, opened 2019-06-26
Remove HAVE_SPL

petk:
The HAVE_SPL symbol is defined in PHP to indicate the presence of the spl extension. Since PHP 5.3 the spl extension is always available and since PHP-7.4 the HAVE_SPL symbol has also been removed.


sirsnyder:
Thx @petk !

Issue #929, opened 2019-02-15
Halting development of pthreads for 7.4


krakjoe:
Hi all,

I'm aware that this is going to annoy some people, so apologies before we begin ...

I do not intend to maintain pthreads anymore, nor encourage it's maintenance: It is, in my opinion, broken by design, and fixing it, is simply not worth it.

The internals world is changing fast and pthreads is not going to be able to fit into that world. It is highly complicated, too difficult to deploy and understand, and slow, necessarily so because of the API it tries to provide. When we get a JIT, we will be in a position where it is much faster to execute your CPU bound code in one normal thread, than distributing that work across many pthreads. There is nothing that can be done to help pthreads while it has the API it has, it's possible that the way pthreads works will actually *prohibit the JIT from compiling code*. Everything about pthreads is putting it in an untenable position for the future.

I've started work on a new extension that will be compatible with the future: https://github.com/krakjoe/parallel

While it is not compatible with pthreads, and will not be, I view it as the only viable replacement for the versions of PHP to come, and my focus has to be on that.

When 7.4 goes into RC stage near the end of this year, this repository will be archived, and no further releases of pthreads will occur on PECL or from this repository.

Thanks for all the effort that contributors have put into pthreads, you've worked hard and I appreciate that, and I would ask that anyone with a focus on pthreads internals today shifts that focus to the future and parallel.

dktapps:
I've been looking at parallel recently and while it looks nice, it can't replace pthreads in its current state due to a few things missing - most notably on my part I need thread-safe queues which parallel doesn't appear to have any way to create.

It appears that parallel would fit in well for a pool and task style architecture in its current form. However, for long-life threads like PocketMine's RakLib library needs, we need thread safe ways to transfer data without closure returns, since there is a persistent context which has to stick around for the duration of the runtime, and also because it's active for the whole time period.

I assume there is more planned for parallel?

rthrfrd:
Thanks and good luck! Limiting closures sounds like a good approach. We applied the same restrictions on them at a PHP level to be able to use them in conjunction with threads - 'compiling' closures into __invokable classes prevented the use of features that made closures un-threadable.

krakjoe:
@dktapps very much unfinished, you can see from commits it has existed for ~four days, although I've been considering it for a long time, I really only just got down to code.

[...]

PR #951, opened 2019-11-06
fix superglobals JIT

dktapps:
Superglobals referenced in copied code are not initialized because they are JIT-initialized when compiled. This can be worked around by setting `auto_globals_jit` to `0` or with this code.

Ideally we would manually JIT these from copied functions like parallel, but this solution is simpler and works well enough as a stopgap for now.

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -756,6 +756,15 @@ int pthreads_prepared_startup(pthreads_object_t* thread, pthreads_monitor_t *rea
 		php_request_startup();
 		PG(during_request_startup) = 0;
 
+		zend_auto_global *auto_global;
+		ZEND_HASH_FOREACH_PTR(CG(auto_globals), auto_global) {
+			if (auto_global->auto_global_callback) {
+				//compiler would normally JIT these globals, but that won't happen for copied code
+				//ideally we would do manual JIT on function copy, but this will work as a stopgap for now
+				auto_global->armed = auto_global->auto_global_callback(auto_global->name);
+			} else auto_global->armed = 0;
+		} ZEND_HASH_FOREACH_END();
+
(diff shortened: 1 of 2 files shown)


sirsnyder:
I'll take a closer look at it the next few days

krakjoe:
#929 

> When 7.4 goes into RC stage near the end of this year, this repository will be archived, and no further releases of pthreads will occur on PECL or from this repository.

RC stage is nearly over, archiving is imminent.

dktapps:
I suspect there are more users than just myself who are not ready to move away from pthreads yet, so I made this PR in case someone may find it useful.

I plan to maintain my fork in the 7.4 period, regardless of JIT availablity.

sirsnyder:
Do you really want to do this? pthreads is such a good extension for php. Through object sharing, pthreads provides the capabilities of other multithreaded languages, such as Java. My applications (including a full runtime environment) are based on this key feature, which otherwise does not provide any extension in the scope. JIT and performance is a nice to have. What good is a JIT when the application stops working? Nothing. The application has to work, then comes the performance. 

You have a lot of projects, I can understand that they are doing a lot of work and you want to focus on parallel for php8 with JIT support. Just give pthreads to Dylan and me and we'll do it. I'm sure we have some good ideas for further development & features. Otherwise, only the fork remains to a new project. Hopefully then again in the PECL.

Thank you @krakjoe

dktapps:
FWIW I am not exactly enamoured of pthreads as it stands, but the migration cost to parallel is just too big, especially considering that parallel has some of its own gremlins to make things confusing (I'm looking at you, object copying).

pthreads hasn't run out of road as far as I'm concerned because it's still in active use by thousands of PM servers worldwide.

